PR #22, opened 2023-08-28
fix: Fix WalletInput component when changing its value programmatically

RuggeroCino:
## Description

- On the delegation flow, we programmatically change the value of the WalletInput component by setting the connected address value. By doing this, the resolved values are not set to undefined as done on the `setValue` callback. This causes an indefinite `resolveValues` loop that set back and forth the old and new addresses when the old address has an ENS domain linked.

## Type of change

## Checklist:

-   [x] I have updated the `CHANGELOG.md` file in the root folder of the package after the [UPCOMING] title and before the latest version.


sepehr2github (approved):
Thanks for the changes

PR #49, opened 2023-11-23
Feature: APP-2612 - Implement Avatar component

Fabricevladimir:
## Description

- adds the avatar component

Task: [APP-2612](https://aragonassociation.atlassian.net/browse/APP-2612)

## Type of change

## Checklist:

-   [x] I have updated the `CHANGELOG.md` file in the root folder of the package after the [UPCOMING] title and before the latest version.

[APP-2612]: https://aragonassociation.atlassian.net/browse/APP-2612?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/avatars/avatar/avatar.tsx
+++ b/src/components/avatars/avatar/avatar.tsx
@@ -96,13 +96,13 @@ export const Avatar: React.FC<IAvatarProps> = (props) => {
 
     // no src or valid image source but broken image
     if (!src || (src && imgErr && isImageSrc(src))) {
-        return <FallbackAvatar className={classNames('bg-neutral-700', containerClasses)} />;
+        return <FallbackAvatar className={classNames('bg-gradients-tl-neutral-900', containerClasses)} />;
     }
 
     // invalid image source ie, label
     if (!isImageSrc(src)) {
         return (
-            <FallbackAvatar className={classNames('bg-primary-700', containerClasses)}>
+            <FallbackAvatar className={classNames('bg-gradients-tl-primary-500-to-800', containerClasses)}>
                 <p className="text-base/tight font-semibold text-neutral-0">{getInitials(src)}</p>
             </FallbackAvatar>
         );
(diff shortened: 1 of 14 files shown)


src/styles/primitives/colors.css
```diff
@@ -67,4 +67,18 @@
...
+    var(--ods-color-primary-500),
+    var(--ods-color-primary-800)
+  );
+
+  --ods-gradient-neutral-900: 31 41 51;
+  --ods-gradient-tl-neutral-900: linear-gradient(
+    135deg,
+    rgb(var(--ods-gradient-neutral-900) / 0.24),
+    rgb(var(--ods-gradient-neutral-900) / 0.96)
+  );
```
Fabricevladimir, line 83: @RuggeroCino, I ran into an interesting little issue where you can either define color stops for the gradients or go all the way and define the whole gradient as a background image. I went for the latter option because it made more sense in terms of on the fly styling, but it does have its drawbacks. Here, for example, I was unable to find a way to add the transparency using the custom ODS color variable as long as that variable was defined as a hex code (no string concatenation in normal css 🤷🏾‍♂️). I'm not at all happy with this solution; what are your thoughts? Right now it's either we define the colors as RGB values or hardcode the gradient stops. Thoughts?

RuggeroCino: I see, CSS custom properties are hard to use with opacity :/ 
I would discuss this with Selim and see if we really need to use opacity for this gradient, if not we could use a lighter neutral color and have something similar to the primary 500 to 800. Otherwise let's discuss about the alternatives

src/components/avatars/avatar/avatar.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+    // valid image src
+    return (
+        <div data-testid="avatar" {...rest} className={containerClasses}>
+            <img src={src} alt="avatar" onError={handleImgError} className="rounded-full object-cover" />
+        </div>
+    );
+};
+
+Avatar.displayName = 'Avatar';
```
RuggeroCino, line 119: The component display name is set automatically, right?

RuggeroCino (changes requested):
Well done @Fabricevladimir! I left some comments

Fabricevladimir:
Reopening new PR to make use of radix-ui's avatar primitive

PR #59, opened 2024-01-08
Feature: APP-2700 - Switch component

Fabricevladimir:
## Description

- Implements the `Switch` component

Task: [APP-2700](https://aragonassociation.atlassian.net/browse/APP-2700)

## Type of change

## Checklist:

        the latest version.

[APP-2700]: https://aragonassociation.atlassian.net/browse/APP-2700?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/switch/switch.stories.tsx
+++ b/src/components/switch/switch.stories.tsx
@@ -1,5 +1,6 @@
 import { useArgs } from '@storybook/preview-api';
 import type { Meta, StoryObj } from '@storybook/react';
+import { useState } from 'react';
 import { Switch } from './switch';
 
 const meta: Meta<typeof Switch> = {
@@ -16,27 +17,46 @@ const meta: Meta<typeof Switch> = {
 
 type Story = StoryObj<typeof Switch>;
 
-export const Default: Story = {
+/**
+ * `Spinner` used as an uncontrolled component
+ */
+export const Uncontrolled: Story = {
+    args: {
+        label: 'Show testnets',
+        name: 'testnet',
+        defaultChecked: true,
+        onCheckedChanged: undefined,
+    },
+};
+
+/**
+ * Controlled usage of the `Spinner` component
+ */
+export const Controlled: Story = {
+    decorators: [
+        (Story) => {
+            const [checked, setChecked] = useState(false);
+
+            const handleCheckedChanged = (newValue: boolean) => {
+                setChecked(newValue);
+            };
+
+            return <Story checked={checked} onCheckedChanged={handleCheckedChanged} />;
+        },
+    ],
+
     args: {
         label: 'Show testnets',
         name: 'testnet',
     },
-};
 
-export const Controlled: Story = {
-    args: {
-        label: 'Show testnets',
-        defaultChecked: true,
-        checked: false,
-    },
     render: function Component(args) {
         const [, setArgs] = useArgs();
 
-        const handleCheckedChanged = (checked: boolean) => {
-            args.onCheckedChanged?.(checked);
-            setArgs({ checked });
+        const handleCheckedChanged = (newValue: boolean) => {
+            args.onCheckedChanged?.(newValue);
+            setArgs({ checked: newValue });
         };
-
         return <Switch {...args} onCheckedChanged={handleCheckedChanged} />;
     },
 };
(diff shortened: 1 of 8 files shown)


Fabricevladimir:
@RuggeroCino, noticed we've got a few components that can be either controlled or uncontrolled, this `Switch` included. I propose we add some sort of warning mechanism so that we can provide proper warnings when the change handler function is missing or a component switches from controlled<=>uncontrolled.  It's not a priority right now, but I think it's a definite nice to have.

Here's an interesting library that tries to mirror what Facebook does with react: https://www.npmjs.com/package/warning.

src/components/switch/switch.tsx
```diff
@@ -0,0 +1,104 @@
...
+    const internalId = useId();
+    const id = propId ?? internalId;
+
+    const switchProps = {
+        id,
+        name,
+        checked,
+        disabled,
+        defaultChecked,
+        'data-testid': 'Switch',
```
RuggeroCino, line 89: We should avoid using the `data-testid` attribute to access elements during tests, this is also discouraged by [Testing Library](https://testing-library.com/docs/queries/bytestid/). It's usually better access elements using their `role`, in this case `getByRole('switch')`

src/components/switch/switch.stories.tsx
```diff
@@ -0,0 +1,64 @@
...
+        },
+    ],
+
+    args: {
+        label: 'Show testnets',
+        name: 'testnet',
+    },
+
+    render: function Component(args) {
+        const [, setArgs] = useArgs();
```
RuggeroCino, line 54: On other stories we implement a component and pass it to the `render` function when the component needs to use hooks (e.g.  useState). Can't we do the same here? Is using `decorators` and `useArgs` better that using a component?

src/components/switch/switch.test.tsx
```diff
@@ -0,0 +1,61 @@
+import { fireEvent, render, screen } from '@testing-library/react';
+import { Switch, type ISwitchProps } from '../switch';
```
RuggeroCino, line 2:
```suggestion
import { Switch, type ISwitchProps } from './switch';
```

RuggeroCino (commented):
Well done @Fabricevladimir 👏🏻 Left some small comments

src/components/switch/switch.stories.tsx
Fabricevladimir: Oh yeah, Storybook hasn't at all been playing nice with me when it comes to controlled components. When I render the function the usual way, the state doesn't get updated in the documentation page even though it works just fine for the individual story.

RuggeroCino: Mm.. could you try something like that:
```
/**
 * Controlled usage of the `Switch` component
 */
const ControlledComponent = (props: ISwitchProps) => {
    const [checked, setChecked] = useState(false);

    return <Switch checked={checked} onCheckedChanged={setChecked} {...props} />;
};

export const Controlled: Story = {
    render: ({ onCheckedChanged, ...props }: ISwitchProps) => <ControlledComponent {...props} />,
    args: {
        label: 'Show testnets',
        name: 'testnet',
    },
};
```

It might be caused by `Storybook` setting a default function for the `onCheckedChanged` property

src/components/switch/switch.tsx
```diff
@@ -0,0 +1,103 @@
...
+    const switchProps = {
+        id,
+        name,
+        checked,
+        disabled,
+        defaultChecked,
+    };
+
+    return (
+        <div className={classNames('inline-flex items-center gap-x-2 md:gap-x-3', className)} {...otherProps}>
```
RuggeroCino, line 92: As of the discussion on Discord, the label color should not get lighter on focus

src/components/switch/switch.stories.tsx
Fabricevladimir: This works halfway. It's not able to capture the action since `onCheckedChanged` is deliberately removed.

RuggeroCino: Yes because we always set it on the `ControlledComponent` component through `onCheckedChanged={setChecked}` and prevent it to be overriden by the Storybook props. The above works on my end, let me know if you have issues and we can have a quick sync

RuggeroCino (approved):
👏🏻 👏🏻 👏🏻

PR #71, opened 2024-01-29
feat: APP-2744 - Implement TextAreaRichText component

RuggeroCino:
## Description

- Implement `TextAreaRichText` component
- Add `wrapperClassName` property to `InputContainer` component to customise the input wrapper
- Update `InputContainer` props to accept any HTML div property and support textarea elements
- Add `border-none` Tailwind CSS utility class

Task: [APP-2744](https://aragonassociation.atlassian.net/browse/APP-2744)

## Type of change

## Checklist:

        the latest version.

[APP-2744]: https://aragonassociation.atlassian.net/browse/APP-2744?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/textAreas/textAreaRichText/textAreaRichTextActions.test.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+    it('renders the set link action instead of the unset one when current active node is not a link', () => {
+        const editor = { isActive: () => false } as unknown as Editor;
+        render(createTestComponent({ editor }));
+        expect(screen.getByTestId(IconType.WYSIWYG_LINK_SET)).toBeInTheDocument();
+        expect(screen.queryByTestId(IconType.WYSIWYG_LINK_UNSET)).not.toBeInTheDocument();
+    });
+
+    it('renders the unset link action instead of the set one when current active node is not a link', () => {
+        const editor = { isActive: () => true } as unknown as Editor;
```
Fabricevladimir, line 41: This is the fun part of testing, isn't it? This tests the unit properly but then the integration of the component into the whole as well needs testing. Sigh

src/components/textAreas/textAreaRichText/textAreaRichTextActions.test.tsx
```diff
@@ -0,0 +1,124 @@
+import { fireEvent, render, screen } from '@testing-library/react';
```
Fabricevladimir, line 1: Recently found out about this new companion library that Kent recomends over using the `fireEvent` functionality: https://testing-library.com/docs/user-event/intro/.

Took a look at the article about why/when to choose one over the other, and it's an interesting read.

src/components/textAreas/textAreaRichText/textAreaRichText.tsx
```diff
@@ -0,0 +1,115 @@
...
+        Placeholder.configure({ placeholder, emptyNodeClass: placeholderClasses, showOnlyWhenEditable: false }),
+        Link,
+    ];
+
+    const editor = useEditor({
+        extensions,
+        content: value,
+        editable: !isDisabled,
+        editorProps: {
+            attributes: {
```
Fabricevladimir, line 59: Can we get the text to wrap horizontally? Right now because it doesn't have a width, we can write forever horizontally. 
In the same vein, do you know in terms of design if we want this to have a fixed height/number of lines like a native textarea or should it grow vertically with the content?

src/components/textAreas/textAreaRichText/textAreaRichText.tsx
```diff
@@ -0,0 +1,115 @@
...
+                class: 'outline-none p-4 prose prose-neutral min-h-[144px] h-full max-w-none',
+                role: 'textbox',
```
Fabricevladimir, line 61: Is it possible to change the prose color to `neutral-600` or do you need to add a custom theme?

Fabricevladimir (commented):
Good work! Left a couple of questions & comments

src/components/input/inputContainer/inputContainer.tsx
```diff
@@ -30,7 +31,7 @@ const variantToClassNames: Record<InputVariant | 'disabled', string[]> = {
  * as `InputText`, `InputNumber` and others. It also manages properties that are shared across all input components,
  * including `label`, `helpText` and more.
  */
-export const InputContainer: React.FC<IInputContainerProps> = (props) => {
+export const InputContainer: React.FC<IInputContainerProps> = forwardRef((props, ref) => {
```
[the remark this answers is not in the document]
RuggeroCino: Agree, I'll use the suggested syntax 👌🏻 
Regarding the `ref`, this `ref` property will be specific to the container of the input and not the input itself. Note that the `useInputProps` will set any `ref` prop to the `inputProps` constant and such props will be set to the input directly (see the `InputDate` component for instance where we forward the ref property to the `input` itself

src/components/textAreas/textAreaRichText/textAreaRichText.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+    // Update editable setting on Tiptap editor on isDisabled property change
+    useEffect(() => {
+        editor?.setEditable(!isDisabled);
+    }, [editor, isDisabled]);
+
+    // Add keydown listener to reset expanded state on ESC key down
+    useEffect(() => {
+        const handleKeyDown = ({ key }: KeyboardEvent) => {
+            if (key === 'Escape') {
```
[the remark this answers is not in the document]
RuggeroCino: It's a feature, not a bug 👀 😂 
Jokes aside, I just wanted to reuse the logic of handling the overflow style on the `document.body` element. I fixed the implementation to simply set `isExpanded` to false on Escape key down and moved that logic on a separate `useEffect`.

src/components/textAreas/textAreaRichText/textAreaRichTextActions.tsx
```diff
@@ -0,0 +1,95 @@
...
+        if (url == null || url === '') {
+            unsetLink();
+
+            return;
+        }
+
+        editor?.chain().focus().extendMarkRange('link').setLink({ href: url }).run();
+    }, [editor, unsetLink]);
+
+    const richTextActions = useMemo(() => {
```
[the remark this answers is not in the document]
RuggeroCino: Good catch, fixed the set/unset handling. Regarding exiting the "link" mode without any content after the link, I'm not sure if this is intended by the library so that you can add the url and then insert the name of the link. In any case the current component on the App behaves the same, I would suggest to further investigate on this if we get such a feedback from design :)

src/components/input/inputContainer/inputContainer.tsx
```diff
@@ -43,23 +44,26 @@ export const InputContainer: React.FC<IInputContainerProps> = (props) => {
...
         className,
         id,
+        wrapperClassName,
+        ...otherProps
     } = props;
 
     const processedVariant = isDisabled ? 'disabled' : variant;
     const containerClasses = classNames(
-        'flex h-12 w-full flex-row items-center', // Layout
+        'flex min-h-12 w-full grow flex-row items-center', // Layout
```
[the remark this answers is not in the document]
RuggeroCino: 100%

src/components/textAreas/textAreaRichText/textAreaRichTextActions.test.tsx
RuggeroCino: I've used `userEvent` before as well, it gives a better abstraction when firing events. We might consider following his suggestion of always using `userEvent` whenever possible and fallback to `fireEvent` only when the event is not supported by the first. Will create a ticket for this 👌🏻

src/components/textAreas/textAreaRichText/textAreaRichText.tsx
RuggeroCino: Fixed first point, will ask Selim for second point for both this and the text-area component 👌🏻

src/components/textAreas/textAreaRichText/textAreaRichText.tsx
RuggeroCino: No not really, what we can do is either choosing a grey scale between `slate`, `gray`, `neutral` or setup our own (https://tailwindcss.com/docs/typography-plugin#adding-custom-color-themes), will check later with Selim

Fabricevladimir: I was afraid of that lol

Fabricevladimir (approved):
Yes! Looking good

PR #72, opened 2024-01-29
Feature: APP-2729 - Implement Input Number component

Fabricevladimir:
## Description

- Implements the `InputNumber` component

Task: [APP-2729](https://aragonassociation.atlassian.net/browse/APP-2729)

## Type of change

## Checklist:

        the latest version.

[APP-2729]: https://aragonassociation.atlassian.net/browse/APP-2729?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/input/inputNumber/index.ts
+++ b/src/components/input/inputNumber/index.ts
@@ -1,0 +1,1 @@
+export { InputNumber, type IInputNumberProps } from './inputNumber';
--- a/src/components/input/inputNumber/inputNumber.test.tsx
+++ b/src/components/input/inputNumber/inputNumber.test.tsx
@@ -1,0 +1,171 @@
+import { fireEvent, render, screen } from '@testing-library/react';
+import { IconType } from '../../icon';
+import * as InputHooks from '../hooks';
+import { InputNumber, type IInputNumberProps } from './inputNumber';
+
+describe('<InputNumber /> component', () => {
+    const createTestComponent = (props?: Partial<IInputNumberProps>) => {
+        const completeProps: IInputNumberProps = {
+            ...props,
+        };
+
+        return <InputNumber {...completeProps} />;
+    };
+
+    it('renders an input with increment and decrement buttons', () => {
+        render(createTestComponent());
+
+        expect(screen.getByRole('textbox')).toBeInTheDocument();
+        expect(screen.getAllByRole('button').length).toEqual(2);
+        expect(screen.getByTestId(IconType.ADD)).toBeInTheDocument();
+        expect(screen.getByTestId(IconType.REMOVE)).toBeInTheDocument();
+    });
+
+    it('renders a disabled input with no spin buttons when isDisabled is set to true', () => {
+        render(createTestComponent({ isDisabled: true }));
+
+        expect(screen.getByRole('textbox')).toBeDisabled();
+        expect(screen.queryAllByRole('button').length).toEqual(0);
+    });
+
+    it('should default step to 1 when given value less than zero', () => {
+        const step = -15;
+        render(createTestComponent({ step }));
+        expect(screen.getByRole('textbox')).toHaveAttribute('step', '1');
+    });
+
+    it('should default step to 1 when given value is zero', () => {
+        const step = 0;
+        render(createTestComponent({ step }));
+        expect(screen.getByRole('textbox')).toHaveAttribute('step', '1');
+    });
+
+    describe('increment button', () => {
+        const useNumberMaskMock = jest.spyOn(InputHooks, 'useNumberMask');
+
+        afterEach(() => {
+            useNumberMaskMock.mockReset();
+        });
+
+        const testIncrementLogic = ({
+            expectedValue,
+            ...props
+        }: Partial<IInputNumberProps> & { expectedValue: string }) => {
+            const setValue = jest.fn();
+            const hookResult = {
+                setValue,
+                value: props.value,
+                unmaskedValue: props.value,
+            } as unknown as InputHooks.IUseNumberMaskResult;
+            useNumberMaskMock.mockReturnValue(hookResult);
+
+            render(createTestComponent({ ...props }));
+
+            const [, incrementButton] = screen.getAllByRole<HTMLButtonElement>('button');
+            fireEvent.click(incrementButton);
+
+            expect(setValue).toHaveBeenCalledWith(expectedValue);
+        };
+
+        it('should increment by one (1) with default parameters', () => {
+            testIncrementLogic({ expectedValue: '1' });
+        });
+
+        it('should return the maximum when the newly generated value exceeds the maximum', () => {
+            const max = '5';
+            const step = '2';
+            const value = '4';
+            testIncrementLogic({ max, step, value, expectedValue: max });
+        });
+
+        it('should increment to the minimum when the newly generated value is less than the minimum', () => {
+            const max = '5';
+            const min = '0';
+            const step = '2';
+            const value = '-7';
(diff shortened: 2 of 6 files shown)


src/components/input/inputNumber/index.ts
RuggeroCino, line 1: Exports are missing on this index file and on the `components/input/index.ts` file

src/components/input/inputNumber/inputNumber.tsx
```diff
@@ -0,0 +1,204 @@
...
+        newValue = Math.max(min, Math.min(max, newValue));
+        setValue(newValue.toString());
+    };
+
+    return (
+        <InputContainer
+            className={classNames('relative', containerClassName)}
+            {...otherContainerProps}
+            isDisabled={isDisabled}
+        >
```
RuggeroCino, line 133: The `relative` class is probably not needed here, right? We can just:
```suggestion
        <InputContainer isDisabled={isDisabled} {...otherContainerProps}>
```

src/components/input/inputNumber/inputNumber.tsx
```diff
@@ -0,0 +1,204 @@
...
+            )}
+            <input
+                ref={numberMaskRef}
+                min={min}
+                max={max}
+                step={step}
+                value={isFocused ? maskedValue : suffixedValue}
+                onBlur={handleBlur}
+                onFocus={handleFocus}
+                disabled={isDisabled}
```
RuggeroCino, line 151: The `disabled` property should already be set on the `otherInputProps`
```suggestion
```

src/components/input/inputNumber/inputNumber.tsx
```diff
@@ -0,0 +1,204 @@
...
+        if (parsedValue > max) {
+            setValue(parsedValue.toString());
+            return;
+        }
+
+        // increment directly to the minimum if value is less than the minimum
+        if (parsedValue < min) {
+            setValue(min.toString());
+            return;
+        }
```
RuggeroCino, line 87: Can the unmasked value be lower or greater than the min / max? Doesn't `react-imask` already restrict this?

src/components/input/inputNumber/inputNumber.tsx
```diff
@@ -0,0 +1,204 @@
...
+
+        // increment value with step if it's smaller than the initial value
+        if (newValue < parsedValue) {
+            newValue += step;
+        }
+
+        // ensure the new value is within the min and max range
+        newValue = Math.max(min, Math.min(max, newValue));
+        setValue(newValue.toString());
```
RuggeroCino, line 103: I would probably simplify this as:
```suggestion
        // Ensure newValue is a multiple of step value
        const newValue = (Math.floor(parsedValue / step) + 1) * step;

        // Ensure newValue is not greater than max value
        setValue(Math.min(max, newValue).toString());
```

src/components/input/inputNumber/inputNumber.tsx
```diff
@@ -0,0 +1,204 @@
...
+ * @param value - The value of the number input. Defaults to '0'.
+ * @param min - The minimum value of the number input. Defaults to Number.MIN_SAFE_INTEGER.
+ * @param max - The maximum value of the number input. Defaults to Number.MAX_SAFE_INTEGER.
+ * @returns  An object with the parsed values.
+ */
+function parseInputs(
+    value: IInputNumberProps['value'] = '',
+    step: IInputNumberProps['step'] = 1,
+    min: IInputNumberProps['min'] = Number.MIN_SAFE_INTEGER,
+    max: IInputNumberProps['max'] = Number.MAX_SAFE_INTEGER,
```
RuggeroCino, line 182: Let's specify these props on the `IInputNumberProps` interface so that we can document their default values. As we're going to specify these props there, I would suggest to only accept the number type and we can remove much of this logic

RuggeroCino (commented):
Thanks for the implementation @Fabricevladimir! Left some comments and questions

src/components/input/inputNumber/inputNumber.tsx
Fabricevladimir: Yup, it should

src/components/input/inputNumber/inputNumber.tsx
Fabricevladimir: A bit iffy about this for a couple of reasons. 
1. It's a major departure from the native html input API; the more familiar we keep the API, the easier the DX and adoption I think
2. `value` has to be at least a `string` (or `string` | `number`) for it to make sense with the IMask `onChange` type which is returning a string. Forcing `value` into a `number` means the burden falls on the consumer to map cast the controlled `onChange` value into a `number` when setting the state.

src/components/input/inputNumber/inputNumber.tsx
RuggeroCino: Okay I see, then we can also remove this check, right?
```
        // increment directly to the minimum if value is less than the minimum
        if (parsedValue < min) {
            setValue(min.toString());
            return;
        }
```

RuggeroCino: Okay I see, then we can also remove this check, right?
```
        // increment directly to the minimum if value is less than the minimum
        if (parsedValue < min) {
            setValue(min.toString());
            return;
        }
```

src/components/input/inputNumber/inputNumber.tsx
RuggeroCino: I agree for the `value` property, the comment is related to the `step`, `min` and `max` properties. I mean, we're already setting some default values which are not set by default, I think we should be fine in only setting these props as number.

src/components/input/inputNumber/inputNumber.tsx
Fabricevladimir: Nope, this is mimicking the behavior of the native HTML5 input `number`. Regardless of the `step` value, as long as the input `value` is not defined, the first increment goes straight to the `min` value

RuggeroCino: Oh I see, makes sense thanks 👌🏻  Can we align the implementation of the `handleDecrement` function to this one so that the implemented logic looks the same?

src/components/input/inputNumber/inputNumber.test.tsx
```diff
@@ -0,0 +1,155 @@
...
+            testDecrementLogic({ step, min, expectedValue: min.toString() });
+        });
+
+        it('should decrement to maximum if the initial value minus the step is greater than the maximum', () => {
+            const value = '100';
+            const step = 2;
+            const min = 1;
+            const max = 50;
+            testDecrementLogic({ value, step, min, max, expectedValue: max.toString() });
+        });
```
RuggeroCino, line 147: We won't have the case where `value` is greater than `max`, right? I think there's a similar test where `value` is less than `min`. I'd suggest to remove these two tests to keep the tests aligned with the logic implemented on the component

src/components/input/inputNumber/inputNumber.tsx
Fabricevladimir: Sure thing 👍🏾

src/components/input/inputNumber/inputNumber.test.tsx
Fabricevladimir: It's just this one. Less than `min` is when `value` is undefined. Good catch

RuggeroCino (approved):
Thanks @Fabricevladimir! ✅

PR #75, opened 2024-01-30
feat: APP-2732 - implement inline ADDON element with variants to inputText

thekidnamedkd:
## Description

implements a left or right addon inline with the input and matching variant styles from it's parent inputContainer. borrows the addition of wrapperClassName from @RuggeroCino work on APP-2744 for richtext to handle overflow-hidden.

Task: [APP-2732](https://aragonassociation.atlassian.net/browse/APP-2732)

## Type of change

## Checklist:

-   [ ] My changes generate no new warnings.
-   [ ] I have updated the `CHANGELOG.md` file in the root folder of the package after the [UPCOMING] title and before
        the latest version.

### DIFF
--- a/src/components/input/inputText/inputText.test.tsx
+++ b/src/components/input/inputText/inputText.test.tsx
@@ -24,4 +24,27 @@ describe('<InputText /> component', () => {
         render(createTestComponent({ inputClassName }));
         expect(screen.getByRole('textbox').className).toContain(inputClassName);
     });
+
+    it('renders with a left addon when addonLeft is provided', () => {
+        const addonLeft = 'Left Addon';
+        render(createTestComponent({ addonLeft }));
+        expect(screen.getByText(addonLeft)).toBeInTheDocument();
+    });
+
+    it('renders with a right addon when addonRight is provided', () => {
+        const addonRight = 'Right Addon';
+        render(createTestComponent({ addonRight }));
+        expect(screen.getByText(addonRight)).toBeInTheDocument();
+    });
+
+    it('does not render both addonLeft and addonRight when both are provided', () => {
+        const addonLeft = 'Left Addon';
+        const addonRight = 'Right Addon';
+        // using unknown as Partial<IInputTextProps> to avoid TS error when forcing both addonLeft and addonRight
+        render(createTestComponent({ addonLeft, addonRight } as unknown as Partial<IInputTextProps>));
+        const leftAddonElement = screen.queryByText(addonLeft);
+        const rightAddonElement = screen.queryByText(addonRight);
+        expect(leftAddonElement).not.toBeInTheDocument();
+        expect(rightAddonElement).not.toBeInTheDocument();
+    });
 });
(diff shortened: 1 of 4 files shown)


src/components/input/inputText/inputText.tsx
```diff
@@ -1,14 +1,44 @@
+import classNames from 'classnames';
 import { useInputProps } from '../hooks';
-import { InputContainer, type IInputComponentProps } from '../inputContainer';
+import { InputContainer, type IInputComponentProps, type InputVariant } from '../inputContainer';
 
-export interface IInputTextProps extends IInputComponentProps {}
+type BaseProps = Omit<IInputComponentProps, 'addonLeft' | 'addonRight'>;
+type NoAddonProps = BaseProps & { addonLeft?: never; addonRight?: never };
+type LeftAddonProps = BaseProps & { addonLeft: string; addonRight?: never };
+type RightAddonProps = BaseProps & { addonLeft?: never; addonRight: string };
```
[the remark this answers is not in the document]
thekidnamedkd: I think this works too. I was trying improve DX with hinting via types, but agree this gets a bit complex and brittle. I just pushed up a change that I think accomplishes the same thing with better readability and usage. @RuggeroCino

src/components/input/inputText/inputText.test.tsx
```diff
@@ -24,4 +25,26 @@ describe('<InputText /> component', () => {
...
+        const addon = 'Left Addon';
+        render(createTestComponent({ addon, addonPos: 'left' }));
+        expect(screen.getByText(addon)).toBeInTheDocument();
+    });
+
+    it('renders with a right addon when addonPos is right and addon is a valid string', () => {
+        const addon = 'Right Addon';
+        render(createTestComponent({ addon, addonPos: 'right' }));
+        expect(screen.getByText(addon)).toBeInTheDocument();
+    });
```
RuggeroCino, line 39: If we use the `order` css property as suggested below, we can just write one test to check that the addon is rendered and we don't need to check for class names

src/components/input/inputText/inputText.api.ts
```diff
@@ -0,0 +1,14 @@
+import type { IInputComponentProps } from '../inputContainer';
+
+export interface IInputTextProps extends IInputComponentProps {
+    /**
+     * @property {string} [addon] - Text to be rendered inside the addon. If empty or undefined, no addon will be rendered.
+     */
```
RuggeroCino, line 6: We don't need the `@property` block tag, the `{string}` type or the name of the property:
```suggestion
    /**
     * Text rendered beside the input field. If empty or undefined, no addon will be rendered.
     */
```

src/components/input/inputText/inputText.api.ts
```diff
@@ -0,0 +1,14 @@
...
+    addon?: string;
+
+    /**
+     * @property {string} [addonPos] - Position of the addon. Must be selected if addon is anything but undefined.
+     * @default 'left'
+     */
```
RuggeroCino, line 12: Same, also second part is not true as we default to "left":
```suggestion
    /**
     * Position of the addon.
     * @default 'left'
     */
```

RuggeroCino (approved):
Thank you for the changes @thekidnamedkd! ✅

PR #86, opened 2024-02-06
Feature: APP-2735 - Implement Radio & RadioGroup Components

Fabricevladimir:
## Description

- Implements the `Radio` & `RadioGroup` components

Task: [APP-2735](https://aragonassociation.atlassian.net/browse/APP-2735)

## Type of change

## Checklist:

        the latest version.

[APP-2735]: https://aragonassociation.atlassian.net/browse/APP-2735?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/index.ts
```diff
@@ -7,6 +7,7 @@ export * from './icon';
 export * from './illustrations';
 export * from './input';
 export * from './progress';
+export * from './radioGroup';
```
RuggeroCino, line 10: Even though a `radio` component won't be used on its own, I would structure these components as:
```
/radioGroup
  /radio
    radio.tsx
    [...]
  /radioGroup
    radioGroup.tsx
    [...]
  /radioCard
    radioCard.tsx
    [...]
```
Just to be consistent with the structure of the other components.

src/components/radioGroup/radioGroup.tsx
```diff
@@ -0,0 +1,49 @@
...
+    /**
+     * Currently selected value in the radio group. Providing this property allows
+     * the component to be controlled.
+     */
+    value?: string;
+    /**
+     *  Callback function that is called when the selected value changes
+     * @param value - new selected radio button value
+     */
+    onValueChange?: (value: string) => void;
```
RuggeroCino, line 33: Can we extend the `RadioGroup.Root` props instead? In that way we don't have to specify most of these properties

src/components/radioGroup/radioGroup.tsx
```diff
@@ -0,0 +1,49 @@
...
+}
+
+/**
+ * `RadioGroup` component
+ */
+export const RadioGroup = forwardRef<HTMLDivElement, IRadioGroupProps>(({ className, ...rest }, ref) => {
+    return (
+        <RadixRadioGroup.Root
+            {...rest}
```
RuggeroCino, line 42: Even if it does not make a difference here, I usually place such spread objects at the end of the props so that users are able to override default properties. Therefore I'm used to see this at the end of the properties list.

src/components/radioGroup/radioGroup.test.tsx
```diff
@@ -0,0 +1,56 @@
+import { fireEvent, render, screen } from '@testing-library/react';
+import { Radio } from './radio';
+import { RadioGroup, type IRadioGroupProps } from './radioGroup';
+
+describe('RadioGroup', () => {
```
RuggeroCino, line 5:
```suggestion
describe('<RadioGroup /> component', () => {
```

src/components/radioGroup/radio/radio.tsx
```diff
@@ -0,0 +1,64 @@
...
+    disabled?: boolean;
+    /**
+     * Value of the radio input
+     */
+    value: string;
+    /**
+     * Variant indicating the label positioning
+     * @default right
+     */
+    variant?: 'left' | 'right';
```
RuggeroCino, line 24: I'm calling this as `labelPosition`, I think it better describe the property. Let me ping Selim to check if he's aligned

src/components/radioGroup/radio/radio.tsx
```diff
@@ -0,0 +1,64 @@
+import * as RadioGroup from '@radix-ui/react-radio-group';
+import classNames from 'classnames';
+import type React from 'react';
+import { useId, type ComponentPropsWithoutRef } from 'react';
+import { Icon, IconType } from '../../icon';
+
+export interface IRadioProps extends ComponentPropsWithoutRef<'button'> {
```
RuggeroCino, line 7: I would probably extend the `RadixGroup.Item` properties as well here

src/components/radioGroup/radio/radio.tsx
```diff
@@ -0,0 +1,64 @@
...
+    );
+
+    const labelClasses = classNames(
+        'text-base leading-tight text-neutral-500', // default
+        'hover:text-neutral-800', // hover
+        'peer-disabled:peer-data-[state=unchecked]:text-neutral-300', // unchecked and disabled
+    );
+
+    return (
+        <div className={classNames('flex items-center gap-2 rounded px-0.5 md:gap-3', className)}>
```
RuggeroCino, line 49: I'd suggest to use a padding left/right on the label instead of a gap so that the space between the indicator and label is also clickable as most of other libraries do. Also other libraries mark the label as clickable

src/components/radioGroup/radioGroup.stories.tsx
```diff
@@ -0,0 +1,62 @@
+import type { Meta, StoryObj } from '@storybook/react';
+import { useState } from 'react';
+import { Radio } from './radio';
+import { RadioGroup, type IRadioGroupProps } from './radioGroup';
+
+const meta: Meta<typeof RadioGroup> = {
+    title: 'components/RadioGroup',
```
RuggeroCino, line 7: We don't usually have a folder which also contain stories (see `alerts` and `avatars`). I'd structure this as follows:
```suggestion
    title: 'components/RadioGroup/RadioGroup',
```

src/components/radioGroup/radio/radio.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+        const completeProps = { label: 'test label', value: 'test value', ...props };
+
+        return (
+            <RadioGroup name="Test Group">
+                <Radio {...completeProps} />;
+            </RadioGroup>
+        );
+    };
+
+    it('should render a label and an unchecked radio button', () => {
```
RuggeroCino, line 17: Just to follow current naming convention on tests:
```suggestion
    it('renders a label and an unchecked radio button', () => {
```
Same for the tests below

src/components/radioGroup/radio/radio.test.tsx
```diff
@@ -0,0 +1,83 @@
+import { fireEvent, render, screen } from '@testing-library/react';
+import { IconType } from '../../icon';
+import { RadioGroup } from '../radioGroup';
+import { Radio, type IRadioProps } from './radio';
+
+describe('Radio', () => {
```
RuggeroCino, line 6:
```suggestion
describe('<Radio /> component', () => {
```

RuggeroCino (commented):
Thanks for the implementation @Fabricevladimir!

src/components/radioGroup/radio/radio.tsx
Fabricevladimir: I don't disagree per se, but in turn we must find a way to document props without directing people to Radix.

RuggeroCino: That's actually a good point, we could update the documentation for such components to say something like: "The component is based on the radix-ui radio implementation [here](link), check the available properties there" or something like that. I'd suggest to create a ticket to check how we can best approach this and update the components props/docs accordingly. What do you think?

Fabricevladimir: Yeah, lets investigate because the closer the documentation is, the better for the consumer. Sucks having to open multiple pages to figure out how a singular component works.

src/components/radioGroup/radio/radio.tsx
Fabricevladimir: Oooh nice catch 😄

src/components/radioGroup/radio/radio.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+        render(createTestComponent({ disabled: true }));
+
+        expect(screen.getByRole('radio')).toBeDisabled();
+    });
+
+    it('sets the radio button value correctly', () => {
+        const label = 'Test label';
+        const value = 'Test value';
+
+        render(createTestComponent({ label, value }));
```
RuggeroCino, line 60: The `label` is not needed for this test, right? 
```suggestion
        const value = 'Test value';

        render(createTestComponent({ value }));
```

src/components/radioGroup/radio/radio.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+    });
+
+    it('sets the label position correctly when variant is left', () => {
+        const label = 'Test label';
+        const value = 'Test value';
+        const variant = 'left';
+
+        render(createTestComponent({ label, value, labelPosition: variant }));
+
+        expect(screen.getByRole('radio')).toHaveClass('order-2');
```
RuggeroCino, line 72:
```suggestion
        const variant = 'left';

        render(createTestComponent({ labelPosition: variant }));

        expect(screen.getByRole('radio')).toHaveClass('order-2');
```

src/components/radioGroup/radio/radio.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+    });
+
+    it('sets the label position to right by default', () => {
+        const label = 'Test label';
+        const value = 'Test value';
+
+        render(createTestComponent({ label, value }));
+
+        expect(screen.getByRole('radio')).not.toHaveClass('order-2');
```
RuggeroCino, line 81:
```suggestion
        render(createTestComponent());

        expect(screen.getByRole('radio')).not.toHaveClass('order-2');
```

RuggeroCino (approved):
All good, only left a couple of suggestions to remove some values for some tests, good job 👏🏻

PR #102, opened 2024-02-21
feat: APP-2912 - Implement DialogAlert component

Fabricevladimir:
## Description

- Implements the `DialogAlert` component
- Fixes the `Icon` styling to prevent shrinking

Task: [APP-2897](https://aragonassociation.atlassian.net/browse/APP-2897)

## Type of change

## Checklist:

        the latest version.

[APP-2897]: https://aragonassociation.atlassian.net/browse/APP-2897?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/dialogs/dialogAlert/dialogAlertContent/dialogAlertContent.stories.tsx
```diff
@@ -0,0 +1,69 @@
...
+        actionLabel: 'Action',
+        cancelLabel: 'Cancel',
+    },
+    render: (props) => (
+        <DialogAlertRoot>
+            <DialogAlertTrigger>Trigger alert</DialogAlertTrigger>
+            <DialogAlertContent {...props}>
+                <p>This is an alert!</p>
+            </DialogAlertContent>
+        </DialogAlertRoot>
```
RuggeroCino, line 35: I think we can improve the components composition by dropping the uncontrolled case (which we never use right now and won't use later). In my opinion currently too much logic is being handled by the `content` component and it could be split instead to `title`, `content` & `footer`. By dropping the uncontrolled case, we could have something like:
```
<DialogAlert.Root>
  <DialogAlert.Title />
  <DialogAlert.Content />
  <DialogAlert.Footer />
</DialogAlert.Root>
```
Let's maybe sync later today on this :)

RuggeroCino (commented):
Just left some initial thoughts about the component

src/components/dialogs/dialogAlert/dialogAlertRoot/dialogAlertRoot.tsx
```diff
@@ -0,0 +1,83 @@
...
+    overlayClassName?: string;
+    /**
+     * The visual style variant of the dialog.
+     * @default info
+     */
+    variant?: DialogAlertVariant;
+    /**
+     * Callback function invoked when the open state of the dialog changes.
+     */
+    onOpenChange?(open: boolean): void;
```
RuggeroCino, line 30: Just for consistency
```suggestion
    onOpenChange?: (open: boolean) => void;
```

src/components/dialogs/dialogAlert/dialogAlertRoot/dialogAlertRoot.tsx
```diff
@@ -0,0 +1,83 @@
+import { Content, Overlay, Portal, Root } from '@radix-ui/react-alert-dialog';
+import classNames from 'classnames';
+import type React from 'react';
+import { createContext, useMemo, type ReactNode } from 'react';
+
+export type DialogAlertVariant = 'critical' | 'info' | 'success' | 'warning';
+
+export interface IDialogAlertRootProps {
+    children?: ReactNode;
```
RuggeroCino, line 9: I know, pretty obvious 😅 
```suggestion
    /**
     * Children of the component.
     */
    children?: ReactNode;
```

src/components/dialogs/dialogAlert/dialogAlertRoot/dialogAlertRoot.tsx
```diff
@@ -0,0 +1,83 @@
...
+                    onOpenAutoFocus={onOpenAutoFocus}
+                >
+                    <DialogAlertContext.Provider value={contextValue}>{children}</DialogAlertContext.Provider>
+                </Content>
+            </Portal>
+        </Root>
+    );
+};
+
+DialogAlertRoot.displayName = 'DialogAlert.Root';
```
RuggeroCino, line 83: This is not needed if we're not using `forwardRef`, right? Same for the other components
```suggestion
```

src/components/dialogs/dialogAlert/dialogAlertRoot/dialogAlertRoot.test.tsx
```diff
@@ -0,0 +1,115 @@
...
+import { DialogAlertFooter, type IDialogAlertFooterProps } from '../dialogAlertFooter';
+import { DialogAlertHeader, type IDialogAlertHeaderProps } from '../dialogAlertHeader';
+import { DialogAlertRoot, type IDialogAlertRootProps } from './dialogAlertRoot';
+
+describe('<DialogAlert.Root/> component', () => {
+    const createTestComponent = (
+        rootProps?: Partial<IDialogAlertRootProps>,
+        headerProps?: Partial<IDialogAlertHeaderProps>,
+        contentProps?: Partial<IDialogAlertContentProps>,
+        footerProps?: Partial<IDialogAlertFooterProps>,
```
RuggeroCino, line 13: I wouldn't render the alert header, content & footer by default but just the eventual children property. In that way, we keep these unit tests independent from the implementation/properties of the alert header, content & footer components.
The `renders the alertdialog with the given content` test can simply check if the given children component get rendered when the dialog is open and the rests regarding the `onOpenChange` callback can be implemented by passing a `<Button />` component as children which triggers the `handleCancelBtnClick` callback on-click

src/components/dialogs/dialogAlert/dialogAlertRoot/dialogAlertRoot.stories.tsx
```diff
@@ -0,0 +1,58 @@
...
+    const handleCloseModal = () => {
+        setOpen(false);
+    };
+
+    return (
+        <>
+            <Button variant="primary" onClick={() => setOpen(true)}>
+                Show DialogAlert
+            </Button>
+            <DialogAlertRoot open={open} onOpenChange={setOpen} {...props}>
```
RuggeroCino, line 36: I would import the full `DialogAlert` component and use it like:
```suggestion
            <DialogAlert.Root open={open} onOpenChange={setOpen} {...props}>
```
In that way, if one day we are able to show the code of storybook components (😭 ), we display the usage by only adding one `DialogAlert` import. Same for the other stories

src/components/dialogs/dialogAlert/dialogAlertHeader/dialogAlertHeader.tsx
```diff
@@ -0,0 +1,55 @@
+import { Title } from '@radix-ui/react-alert-dialog';
+import classNames from 'classnames';
+import type React from 'react';
+import { useContext, type ComponentPropsWithoutRef } from 'react';
+import { AvatarIcon } from '../../../avatars';
+import { IconType } from '../../../icon';
+import { DialogAlertContext, type DialogAlertVariant } from '../dialogAlertRoot/dialogAlertRoot';
```
RuggeroCino, line 7:
```suggestion
import { DialogAlertContext, type DialogAlertVariant } from '../dialogAlertRoot';
```

src/components/dialogs/dialogAlert/dialogAlertFooter/dialogAlertFooter.tsx
```diff
@@ -0,0 +1,57 @@
...
+import type React from 'react';
+import { useContext } from 'react';
+import { Button, type IButtonProps } from '../../../button';
+import { DialogAlertContext } from '../dialogAlertRoot';
+
+export interface IDialogAlertFooterProps {
+    /**
+     * Alert dialog primary action button
+     */
+    actionButton: IButtonProps & { label: string };
```
RuggeroCino, line 12: Should we omit the `variant` & `size` properties? In any case, I would suggest to create a new interface like:
```
IDialogAlertFooterAction extends IButtonProps {
  /**
   * [...]
   */
  label: string;

```

src/components/dialogs/dialogAlert/dialogAlertContent/dialogAlertContent.tsx
```diff
@@ -0,0 +1,14 @@
...
+import type React from 'react';
+import { type ComponentPropsWithoutRef } from 'react';
+
+export interface IDialogAlertContentProps extends ComponentPropsWithoutRef<'div'> {}
+
+/**
+ * `DialogAlert.Content` component.
+ */
+export const DialogAlertContent: React.FC<IDialogAlertContentProps> = ({ className, ...otherProps }) => {
+    return <div className={classNames('overflow-auto px-4 md:px-6', className)} {...otherProps} />;
```
RuggeroCino, line 11: Don't we need to add a background color to the dialog content/header/footers?
We should also increase the z-index of the overlay/dialog-root
Screenshot 2024-02-23 at 11 29 16

src/components/dialogs/dialogAlert/dialogAlertRoot/dialogAlertRoot.tsx
```diff
@@ -0,0 +1,83 @@
...
+export interface IDialogAlertContext {
+    variant: DialogAlertVariant;
+}
+
+export const DialogAlertContext = createContext<IDialogAlertContext>({ variant: 'info' });
+
+/**
+ * `DialogAlert.Root` component.
+ */
+export const DialogAlertRoot: React.FC<IDialogAlertRootProps> = (props) => {
```
RuggeroCino, line 50: Currently the dialog does not close automatically on esc keydown, is this intended?

src/components/dialogs/dialogAlert/dialogAlertContent/dialogAlertContent.tsx
RuggeroCino, line 11: I saw that on Figma the vertical padding is applied to a inner div, I think that the `content` should also have a default vertical padding. What do you think?

RuggeroCino (commented):
Good stuff 👏🏻

src/components/dialogs/dialogAlert/dialogAlertRoot/dialogAlertRoot.tsx
Fabricevladimir: Ah, yes. I didn't catch that. Because the dialog is controlled, this has to be handled manually. Good catch

src/components/dialogs/dialogAlert/dialogAlertRoot/dialogAlertRoot.tsx
Fabricevladimir: It's not needed, agreed, but if we're going to display them as `Component.Root` on Storybook, I think it makes sense to align the display name as well.
What do you think?

src/components/dialogs/dialogAlert/dialogAlertRoot/dialogAlertRoot.tsx
Fabricevladimir: Good catch, was a typo.

src/components/dialogs/dialogAlert/dialogAlertContent/dialogAlertContent.tsx
Fabricevladimir: The header and footer has the gradient background, remember?

Fabricevladimir: Per designs, only horizondal padding should be added. Vertical padding is left to the content to define.

CHANGELOG.md
```diff
@@ -16,6 +20,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 -   Rename `isDisabled` property of input components to `disabled`
 -   Bump `ip` library from 2.0.0 to 2.0.1
 
+### Fixed
+
+-   `Icon` styling to prevent shrinking
+
```
cgero-eth, line 26: This should be under the `unreleased` section

src/components/dialogs/dialogAlert/dialogAlertRoot/dialogAlertRoot.tsx
cgero-eth: Oh I see, I wouldn't add code to the component because of Storybook, we anyway need to figure out how to properly display the code on storybook

src/components/dialogs/dialogAlert/dialogAlertRoot/dialogAlertRoot.test.tsx
```diff
@@ -0,0 +1,40 @@
+import { render, screen } from '@testing-library/react';
+import { userEvent } from '@testing-library/user-event';
+import { type ReactNode } from 'react';
+import { DialogAlertRoot, type IDialogAlertRootProps } from './dialogAlertRoot';
+
+describe('<DialogAlert.Root/> component', () => {
+    const createTestComponent = (rootProps?: Partial<IDialogAlertRootProps>, children?: ReactNode) => {
```
cgero-eth, line 7: The root component already accept the `children` property, right? We can pass it as props
```suggestion
    const createTestComponent = (props?: Partial<IDialogAlertRootProps>) => {
```

src/components/dialogs/dialogAlert/dialogAlertHeader/dialogAlertHeader.tsx
```diff
@@ -0,0 +1,55 @@
...
+    warning: 'text-warning-600',
+};
+
+/**
+ * `DialogAlert.Header` component
+ *
+ * **NOTE**: This component must be used inside a `<DialogAlert.Root />` component.
+ */
+export const DialogAlertHeader: React.FC<IDialogAlertHeaderProps> = ({ title }) => {
+    const { variant } = useContext(DialogAlertContext);
```
cgero-eth, line 36: We need to pass all the `otherProps` object to the main div 
```suggestion
export const DialogAlertHeader: React.FC<IDialogAlertHeaderProps> = (props) => {
    const { title, ...otherProps } = props;
    const { variant } = useContext(DialogAlertContext);
```

src/components/dialogs/dialogAlert/index.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const DialogAlert = {
+    Content: DialogAlertContent,
+    Footer: DialogAlertFooter,
+    Header: DialogAlertHeader,
+    Root: DialogAlertRoot,
+};
+
+export { type IDialogAlertContentProps } from './dialogAlertContent';
+export { IDialogFooterAction, type IDialogAlertFooterProps } from './dialogAlertFooter';
```
cgero-eth, line 14:
```suggestion
export { IDialogFooterAction, type IDialogAlertFooterAction, type IDialogAlertFooterProps } from './dialogAlertFooter';
```

src/components/dialogs/dialogAlert/dialogAlertFooter/dialogAlertFooter.tsx
```diff
@@ -0,0 +1,68 @@
+import { Action, Cancel } from '@radix-ui/react-alert-dialog';
+import classNames from 'classnames';
+import type React from 'react';
+import { useContext, type AnchorHTMLAttributes, type ButtonHTMLAttributes } from 'react';
+import { Button, type IButtonBaseProps } from '../../../button';
+import { DialogAlertContext } from '../dialogAlertRoot';
+
+export type IDialogFooterAction = (
```
cgero-eth, line 8:
```suggestion
export type IDialogAlertFooterAction = (
```

cgero-eth (commented):
Just some minor comments, PR looks good 👌🏻

Fabricevladimir:
> Oh I see, I wouldn't add code to the component because of Storybook, we anyway need to figure out how to properly display the code on storybook

The code isn't being added _because_ of Storybook. I think it's a cleaner/ more readable tree rendering it this way because the component has a lot of parts with starting with the same `DialogAlert` prefix in their display names.

cgero-eth (approved):
👏🏻👏🏻👏🏻

PR #113, opened 2024-02-28
feat: Update @typescript-eslint and eslint-plugin-storybook dependencies

cgero-eth:
## Description

- Update @typescript-eslint and eslint-plugin-storybook dependencies
- Set minimum required node version to 18.18.0 (required by `@typescript-eslint` v7)

## Type of change

## Checklist:

        the latest version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,8 +93,8 @@
     "@testing-library/react": "^14.2.1",
     "@testing-library/user-event": "^14.5.2",
     "@types/jest": "^29.5.12",
-    "@typescript-eslint/eslint-plugin": "^6.21.0",
-    "@typescript-eslint/parser": "^6.21.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -126,7 +126,7 @@
   },
   "homepage": "https://github.com/aragon/ods#readme",
   "engines": {
-    "node": ">=18",
+    "node": ">=18.18.0",
```
Fabricevladimir, line 129: Hmm, we should check over on the App if everything works as expected with new Node version

Fabricevladimir (approved):
Thanks, @cgero-eth

package.json
cgero-eth: Yes, we should update the app min node version to 18.18 as well. Given that it is not a major version everything should work fine 👍🏻

PR #117, opened 2024-03-04
feat: APP-2939 - Implement DaoAvatar module component

Fabricevladimir:
## Description

- Implements `DaoAvatar` module component

Task: [APP-2939](https://aragonassociation.atlassian.net/browse/APP-2939)

## Type of change

## Checklist:

        the latest version.

[APP-2939]: https://aragonassociation.atlassian.net/browse/APP-2939?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/components/dao/daoAvatar/daoAvatar.tsx
+++ b/src/modules/components/dao/daoAvatar/daoAvatar.tsx
@@ -6,12 +6,12 @@ export interface IDaoAvatarProps extends Omit<IAvatarProps, 'fallback'> {
     /**
      * Name of the DAO
      */
-    daoName: string;
+    name?: string;
 }
 
 export const DaoAvatar: React.FC<IDaoAvatarProps> = (props) => {
-    const { daoName, size = 'lg', className, ...otherProps } = props;
-    const daoInitials = getDaoInitials(daoName).toUpperCase();
+    const { name, size = 'lg', className, ...otherProps } = props;
+    const daoInitials = getDaoInitials(name).toUpperCase();
 
     return (
         <Avatar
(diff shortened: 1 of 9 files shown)


src/modules/components/dao/daoAvatar/daoAvatar.tsx
```diff
@@ -0,0 +1,57 @@
+import classNames from 'classnames';
+import type React from 'react';
+import { Avatar, type IAvatarProps } from '../../../../core';
+
+export interface IDaoAvatarProps extends Omit<IAvatarProps, 'fallback'> {
+    /**
+     * Name of the DAO
+     */
+    daoName: string;
```
cgero-eth, line 9: Given that we're already in the dao context, I'd suggest to simply name the property as `name`. Similar for the DaoListItem properties (name, description, logo, ..)
```suggestion
    name: string;
```

src/modules/components/dao/daoAvatar/daoAvatar.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Get the initials for a DAO
+ * @param daoName name of the DAO
+ * @returns the DAO initials
+ */
+function getDaoInitials(daoName: IDaoAvatarProps['daoName']): string {
+    if (!daoName) {
+        return '';
+    }
```
cgero-eth, line 43: The `daoName` can't be undefined and the empty string case is already handle below, right?
```suggestion
```

cgero-eth (commented):
Thanks @Fabricevladimir

src/modules/components/dao/daoAvatar/daoAvatar.tsx
Fabricevladimir: `daoName` shouldn't be undefined, but it certainly can be. The DAO name isn't on chain but pinned wherever, so the structure cannot be 100% enforced. Sure the consumer can pass a default empty string, but this ensures the whole application doesn't crash since this string is being processed by the component.

Fabricevladimir: Probably the most complete solution, yes.

cgero-eth (approved):
LGTM, thanks @Fabricevladimir

PR #122, opened 2024-03-12
feat: APP-2941 - Implement ProposalDataListItem module component

Fabricevladimir:
## Description

- Implements the `ProposalDataListItem` module component
- Updates the `Tag` core component primary variant styling

Note: ~~The `Pulse` component will follow in a  separate PR~~ 
Update: `StatePingAnimation` component merged - #126

Task: [APP-2941](https://aragonassociation.atlassian.net/browse/APP-2941)

## Type of change

## Checklist:

        the latest version.

[APP-2941]: https://aragonassociation.atlassian.net/browse/APP-2941?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+        resultComponent = (
+            <ProposalDataListItemResult
+                option={option}
+                voteAmount={voteAmount}
+                votePercentage={votePercentage}
+                type={type}
+            />
+        );
+    }
```
cgero-eth, line 48: I would change the component props interface a bit to avoid such logic in place, when I see `let` initialisations inside a component I usually think that the code can be improved. 
Given that we already have a type property which tells us what to expect on the properties, we could have something like:
```
export interface IProposalListItemBaseProps<TType extends ProposalType = 'majorityVoting'>
    extends IDataListItemProps,
        IWeb3ComponentProps {
    [...]

    type: TType;
    result: TType extends 'majorityVoting' ? IMajorityVotingResult : IApprovalThresholdResult;
}
```
With this we won't have to destructure the props twice (`nonWeb3Props` & `otherProps`) and we could just pass to the result component the `result` and `type` properties. What do you think?

src/modules/components/proposal/proposalDataListItem/proposalDataListItemResult/proposalDataListItemResult.tsx
```diff
@@ -0,0 +1,53 @@
+import { Progress, formatterUtils } from '../../../../../core';
+import { type IApprovalThresholdResult, type IMajorityVotingResult } from '../proposalDataListItemStructure';
+
+export type IProposalDataListItemResultProps = IApprovalThresholdResult | IMajorityVotingResult;
+
+/**
+ * `ProposalDataListItemResult` component
+ */
+export const ProposalDataListItemResult: React.FC<IProposalDataListItemResultProps> = (props) => {
```
cgero-eth, line 9: Ngl, these two returns inside if statements looks pretty ugly 😅 On one end we could split this into two components, this will make the component easier to test and more scalable when adding more result types. On the other end it is pretty small right now and adding more folders/components might be overkill :/

cgero-eth (commented):
Thanks for the implementation @Fabricevladimir! Left some initial thoughts, I think that the component props can be improved a bit

src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+            {ongoing && resultComponent}
+            <div className="flex items-center gap-x-4 md:gap-x-6">
+                <div className="flex min-h-5 flex-1 text-sm leading-tight text-neutral-600 md:min-h-6 md:text-base">
+                    {/* TODO: apply internationalization [APP-2627] */}
+                    By
+                    {/* using a span here instead of an actual button so that the HTML stays valid 
+                    since the DataList.Item is an anchor tag; must handle set role and handle appropriate keydown event */}
+                    <span
+                        role="button"
```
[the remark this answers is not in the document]
Fabricevladimir: This is indeed a requirement. Clicking on it opens the publisher's profile page on the DAO where the proposal was created. It should technically be a link, but given the parent component is a link, it's either this or disabling the clicking.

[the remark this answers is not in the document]
cgero-eth: I see, found this interesting article explaining an hacky way to have nested links => https://kizu.dev/nested-links/
What do you think? In that way we render the proper anchor tag that gives the accessibility features by default

src/modules/components/proposal/proposalDataListItem/approvalThresholdResult/approvalThresholdResult.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { type IApprovalThresholdResult } from '../proposalDataListItemStructure';
+
+export interface IApprovalThresholdResultProps extends IApprovalThresholdResult {}
+
+/**
+ * `ApprovalThresholdResult` component
+ */
+export const ApprovalThresholdResult: React.FC<IApprovalThresholdResultProps> = (props) => {
+    const { approvalAmount, approvalThreshold } = props;
+    const percentage = Math.min((approvalAmount / approvalThreshold) * 100, 100);
```
cgero-eth, line 11: The `Progress` component already limits the value to be 0 < value < 100, I'd remove the `Math.min` here.
I would probably add a check for approvalThreshold to be !== 0

src/modules/components/proposal/proposalDataListItem/approvalThresholdResult/approvalThresholdResult.test.tsx
```diff
@@ -0,0 +1,27 @@
...
+        const completeProps: IApprovalThresholdResultProps = {
+            approvalAmount: 1,
+            approvalThreshold: 2,
+            ...props,
+        };
+
+        return <ApprovalThresholdResult {...completeProps} />;
+    };
+
+    it('renders the given approval threshold, approval amount and corresponding progressbar', () => {
```
cgero-eth, line 15: It would be nice to test that the amounts are properly formatted and the progress value is correct. The progress value could be checked with `screen.getByRole('progressbar').getAttribute('data-value'))`

src/modules/components/proposal/proposalDataListItem/majorityVotingResult/majorityVotingResult.test.tsx
```diff
@@ -0,0 +1,30 @@
+import { render, screen } from '@testing-library/react';
+import { MajorityVotingResult, type IMajorityVotingResultProps } from './majorityVotingResult';
+
+describe('<ApprovalThresholdResult /> component', () => {
```
cgero-eth, line 4:
```suggestion
describe('<MajorityVotingResult /> component', () => {
```

src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.api.ts
```diff
@@ -0,0 +1,90 @@
...
+     * Number of votes for the winning option
+     */
+    voteAmount: string;
+    /**
+     * Percentage of votes for the winning option
+     */
+    votePercentage: number;
+}
+
+export interface IProposalDataListItemStructureProps extends IProposalListItemBaseProps<ProposalType> {}
```
cgero-eth, line 90: Can't we just rename `IProposalListItemBaseProps` to `IProposalDataListItemStructureProps`?

src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.stories.tsx
```diff
@@ -0,0 +1,83 @@
...
+export const MajorityVoting: Story = {
+    args: {
+        ...baseArgs,
+        type: 'majorityVoting',
+        option: 'yes',
+        voteAmount: '100k wAnt',
+        votePercentage: 15,
+    },
+    render: (props) => (
+        <OdsModulesProvider>
```
cgero-eth, line 51: The `OdsModulesProvider` is already rendered for each story
```suggestion
```

cgero-eth (commented):
Had a look at the other files, looks solid 💪🏻

src/modules/components/proposal/proposalDataListItem/approvalThresholdResult/approvalThresholdResult.tsx
Fabricevladimir: I don't think it's possible to have a zero approvalThreshold. Also, if you've a proposal without people to vote on, this component crashing is the least of your issues lol

src/modules/components/proposal/proposalDataListItem/approvalThresholdResult/approvalThresholdResult.test.tsx
Fabricevladimir: The progressBar assertion makes sense but not the formatted one. It's in essence just checking if the formatter was used as opposed to whether it's properly formatted.

cgero-eth: Well yes, using the formatter to format the numbers is part of the component's logic, we could add some tests to make sure that the formatter is being used with the correct format. In theory you could assert this through a spy and then `.toHaveBeenCalledWith`, but it's just simpler and more straight forward to assert that a text is rendered on `screen`

CHANGELOG.md
```diff
@@ -36,6 +37,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 -   Set minimum required node version to 18.18.0 (required by `@typescript-eslint` v7)
 -   Update documentation on modules components
 -   Update required dependencies, move `react` and `react-dom` to peer dependencies
+-   Update `Tag` component primary variant styling
```
cgero-eth, line 40: This should be under the `unreleased` header.. We should think about using a tool like [changesets](https://github.com/changesets/changesets) to avoid all these conflicts/issues with the changelog 😅

src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.api.ts
cgero-eth: Oh I see, I thought that with the generic we would get the correct `result` type when checking the `type` property. Maybe doing something similar you did before would be actually better, something like:
```
export interface IProposalDataListItemStructureMajorityVotingProps extends IProposalDataListItemStructureBaseProps {
    type: 'majorityVoting';
    result: IMajorityVotingResult;
}

export interface IProposalDataListItemStructureApprovalThresholdProps extends IProposalDataListItemStructureBaseProps {
    type: 'approvalThreshold';
    result: IApprovalThresholdResult;
}

export type IProposalDataListItemStructureProps =
    | IProposalDataListItemStructureMajorityVotingProps
    | IProposalDataListItemStructureApprovalThresholdProps;
```

src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.tsx
```diff
@@ -0,0 +1,83 @@
...
+            {ongoing && type === 'majorityVoting' && <MajorityVotingResult {...(result as IMajorityVotingResult)} />}
+
+            <div className="flex items-center gap-x-4 md:gap-x-6">
+                <div className="flex min-h-5 flex-1 items-center gap-x-0.5 text-sm leading-tight text-neutral-600 md:min-h-6 md:text-base">
+                    {/* TODO: apply internationalization [APP-2627] */}
+                    By
+                    {/* using solution from https://kizu.dev/nested-links/ to nest anchor tags */}
+                    <object type="disregardType" className="ml-0.5 md:ml-1">
+                        <a
+                            tabIndex={0}
```
cgero-eth, line 64: Can't we use a link component here?

src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+    const ongoing = status === 'active' || status === 'challenged' || status === 'vetoed';
+
+    const publisherIsConnected = isConnected && connectedAddress?.toLowerCase() === publisher.address?.toLowerCase();
+    const publisherLabel = publisherIsConnected
+        ? 'You'
+        : publisher.name ?? addressUtils.shortenAddress(publisher.address as string);
+
+    return (
+        <DataList.Item className="space-y-4" {...otherProps}>
```
cgero-eth, line 44: We probably need to use `classNames`, otherwise this class will be replaced when adding custom classes as prop
```suggestion
        <DataList.Item className={classNames('space-y-4', className)} {...otherProps}>
```

CHANGELOG.md
Fabricevladimir: Damn. Yeah, let me look at that because it's rarely the case that we don't run into a changelog issue, and if you're not careful, stuff will just sneak in. Good catch.

src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.tsx
```diff
@@ -0,0 +1,83 @@
...
+            <ProposalDataListItemStatus date={date} status={status} voted={voted} />
+            <div className="space-y-1">
```
cgero-eth, line 46: I started noticing these `space-*` classes on other components as well, I would suggest to always prefer `gap-*` to set the spacing between the components

cgero-eth (commented):
Last minor discussions, looks good overall 👌🏻

src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.tsx
Fabricevladimir: What's the reasoning behind that? Adding gap means we have to tranform this into either a grid or a flexbox.

src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.tsx
Fabricevladimir: Good catch

src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.tsx
cgero-eth: In general I think that the gap property better represent the spacing idea between elements and it actually applies this in flex/grid containers. `space-*` instead applies margins on elements making sure that such margins are not applied on the first/last elements, sounds a bit more convoluted as idea

Fabricevladimir: Prior to flexbox, that was _the only_ way to do it. I want you to know that I looked up (and couldn't find) an "are you serious" emoji 😅

src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.api.ts
Fabricevladimir: I'll give it a go later and see what I find 👍🏾

src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.tsx
cgero-eth: We should create such emoji :D

src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.api.ts
Fabricevladimir: I think we found a winner!

cgero-eth (approved):
Well done, looks good to me 👏🏻 🚀

PR #130, opened 2024-03-25
feat: APP-2797 - Implement TransactionDataListItem.Structure module component

thekidnamedkd:
## Description

Implement TransactionDataListItemStructure component + timestamp formatter & block explorer link creation utilities 

Task: [APP-2797](https://aragonassociation.atlassian.net/browse/APP-2797)

## Type of change

## Checklist:

        the latest version.

[APP-2797]: https://aragonassociation.atlassian.net/browse/APP-2797?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/modules/components/transaction/transactionDataListItem/transactionDataListItemStructure/transactionDataListItemStructure.tsx
```diff
@@ -0,0 +1,109 @@
...
+        return (
+            <div className="flex size-6 shrink-0 items-center justify-center md:size-8">
+                <Spinner className="transition" variant="neutral" responsiveSize={{ md: 'lg' }} />
+            </div>
+        );
+    };
+
+    return (
+        <DataList.Item
+            className="min-w-fit !py-0 px-4 md:px-6"
```
Fabricevladimir, line 68: The className specified here will be overwritten by the otherProps being spread afterwards. You can use the classNames function to combine that and the className coming from the props. Also, by doing so, you shouldn't neeed to set the padding as important.

src/modules/components/transaction/transactionDataListItem/transactionDataListItemStructure/transactionDataListItemStructure.tsx
```diff
@@ -0,0 +1,109 @@
...
+            {...otherProps}
```
Fabricevladimir, line 69: Spread this last so that the user can override properties like the `href` and `target` if necessary.

src/modules/components/transaction/transactionDataListItem/transactionDataListItemStructure/transactionDataListItemStructure.tsx
```diff
@@ -0,0 +1,109 @@
...
+    } = props;
+
+    const getEffectiveStatus = () => {
+        const type = txType;
+        if (txStatus === TxStatusCode.FAILED) {
+            return {
+                icon: IconType.CLOSE,
+                variant: 'critical' as AvatarIconVariant,
+                heading: 'Failed transaction',
+            };
```
Fabricevladimir, line 43: Why not add the heading, icon, and variant to the `txHeadingStringList`,`txIconTypeList`, and `txVariantList` respectively? That would make it so that you don't need this `getEffectiveStatus` function.

src/modules/components/transaction/transactionDataListItem/transactionDataListItemStructure/transactionDataListItemStructure.tsx
```diff
@@ -0,0 +1,109 @@
...
+            className="min-w-fit !py-0 px-4 md:px-6"
+            {...otherProps}
+            href={createBlockExplorerLink({ chainId, queryType: QueryType.TX, txHash })}
+            target="_blank"
+        >
+            <div className="flex w-full justify-between py-3 md:py-4">
+                <div className="flex items-center gap-x-3 md:gap-x-4">
+                    {getTxIcon()}
+                    <div className="flex w-full flex-col items-start gap-y-0.5">
+                        <Heading size="h5" as="h2">
```
Fabricevladimir, line 77: Are these headings actually defined in the designs or are they approximations?

src/modules/components/transaction/transactionDataListItem/transactionDataListItemStructure/transactionDataListItemStructure.tsx
```diff
@@ -0,0 +1,109 @@
...
+                            {heading}
+                        </Heading>
+                        <Heading className="!text-neutral-500" size="h5" as="h2">
```
Fabricevladimir, line 80: ! often smells! See previous comment

src/modules/components/transaction/transactionDataListItem/transactionDataListItemStructure/transactionDataListItemStructure.tsx
```diff
@@ -0,0 +1,109 @@
...
+                <div className="flex flex-col items-end gap-y-0.5">
+                    <Heading size="h5" as="h2">
+                        {tokenValue && txType !== TransactionType.ACTION ? (
+                            <>
+                                {formatterUtils.formatNumber(tokenValue, { format: NumberFormat.TOKEN_AMOUNT_SHORT })}
+                                {` ${tokenSymbol}`}
+                            </>
+                        ) : (
+                            `-`
+                        )}
```
Fabricevladimir, line 95: Mind breaking this up into two explicit conditions? It's often clearer and cleaner to avoid ternary in JSX

src/modules/components/transaction/transactionDataListItem/transactionDataListItemStructure/transactionDataListItemStructure.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+    const { icon, variant, heading } = getEffectiveStatus();
+
+    const getTxIcon = () => {
+        if (txStatus !== TxStatusCode.PENDING) {
+            return <AvatarIcon className="shrink-0" variant={variant} icon={icon} responsiveSize={{ md: 'md' }} />;
+        }
+        return (
+            <div className="flex size-6 shrink-0 items-center justify-center md:size-8">
+                <Spinner className="transition" variant="neutral" responsiveSize={{ md: 'lg' }} />
```
Fabricevladimir, line 61: Something is up with the Spinner color. I think the base component might have been updated. I would suggest maybe tackling this as your next task if it's not a quick fix.

src/modules/components/transaction/transactionDataListItem/transactionDataListItemStructure/transactionDataListItemStructure.tsx
```diff
@@ -0,0 +1,109 @@
...
+    };
+
+    return (
+        <DataList.Item
+            className="min-w-fit !py-0 px-4 md:px-6"
+            {...otherProps}
+            href={createBlockExplorerLink({ chainId, queryType: QueryType.TX, txHash })}
+            target="_blank"
+        >
+            <div className="flex w-full justify-between py-3 md:py-4">
```
Fabricevladimir, line 73: This isn't really space between as much as gap-y-3 and gap-y-4 (on md). Of course visually it works, but I'm highlighting in case it's relevant when truncating on small screens. The date, token value and symbol are wrapping on smaller screens

src/modules/components/transaction/transactionDataListItem/transactionDataListItemStructure/transactionDataListItemStructure.stories.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { TransactionDataListItemStructure } from './transactionDataListItemStructure';
+
+const meta: Meta<typeof TransactionDataListItemStructure> = {
+    title: 'Modules/Components/Transaction/TransactionDataListItem.Structure',
+    component: TransactionDataListItemStructure,
+    tags: ['autodocs'],
+    parameters: {
+        design: {
+            type: 'figma',
+            url: 'https://www.figma.com/file/ISSDryshtEpB7SUSdNqAcw/branch/P0GeJKqILL7UXvaqu5Jj7V/Aragon-ODS?type=design&node-id=14385%3A30819&mode=dev',
```
Fabricevladimir, line 12:
```suggestion
            url: 'https://www.figma.com/file/P0GeJKqILL7UXvaqu5Jj7V/v1.1.0?type=design&node-id=445-5113&mode=design&t=qzF3muTU7z33q8EX-4',
```

src/modules/components/transaction/transactionDataListItem/transactionDataListItemStructure/transactionDataListItemStructure.stories.tsx
```diff
@@ -0,0 +1,37 @@
...
+        },
+    },
+};
+
+type Story = StoryObj<typeof TransactionDataListItemStructure>;
+
+/**
+ * Default usage example of the TransactionDataList module component.
+ */
+export const Default: Story = {
```
Fabricevladimir, line 27: Annoying, but I think given the different types of actions and the their statuses, it would make sense to have a couple corresponding stories to highlight the usage, at least one for smart contract action and one for normal withdraw or deposit.

src/modules/components/transaction/transactionDataListItem/transactionDataListItemStructure/transactionDataListItemStructure.api.ts
```diff
@@ -0,0 +1,71 @@
...
+     */
+    tokenSymbol?: string;
+    /**
+     * The token value in the transaction.
+     */
+    tokenValue?: number;
+    /**
+     * The type of transaction.
+     */
+    txType?: TransactionType;
```
Fabricevladimir, line 54: Why do we want this to be optional? Same question for the `txStatus` and `txHash`

src/modules/components/transaction/transactionDataListItem/transactionDataListItemStructure/transactionDataListItemStructure.api.ts
```diff
@@ -0,0 +1,71 @@
...
+    /**
+     * The network state of the transaction.
```
Fabricevladimir, line 56: I think this might be slightly ambiguous. What do you think of something like "represents the current status of a blockchain transaction"?

src/modules/components/transaction/transactionDataListItem/transactionDataListItemStructure/transactionDataListItem.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+        render(createTestComponent({ txType }));
+        const transactionTypeHeading = screen.getByText('Smart contract action');
+        expect(transactionTypeHeading).toBeInTheDocument();
+    });
+
+    // fails on the local run, but passes on the CI for relative unix timestamps (CI server location == UTC 0)
+    it('renders the formatted date', () => {
+        const unixTimestamp = 1628841600;
+        render(createTestComponent({ unixTimestamp }));
+        const formattedDate = screen.getByText('January 19, 1970 at 8:27 PM');
```
Fabricevladimir, line 32: Shouldn't using the formatter in the test actually fix the local and CI time issue?
```suggestion
        const formattedDate = screen.getByText(formatDate(unixTimeStamp));
```
With this, you are just testting that the date is being formatted. You needn't worry about the actual format and value; those will be tested with the date formatter util's own unit tests.

That said, take a look at my comment about the formatter itself before you make any changes.

src/modules/components/transaction/transactionDataListItem/transactionDataListItemStructure/transactionDataListItem.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+        const tokenValue = 10;
+        render(createTestComponent({ tokenSymbol, tokenValue }));
+        const tokenPrintout = screen.getByText('10 ETH');
+        expect(tokenPrintout).toBeInTheDocument();
+    });
+
+    it('renders the formatted USD estimate', () => {
+        const usdEstimate = 100;
+        render(createTestComponent({ usdEstimate }));
+        const formattedUsdEstimate = screen.getByText('$100.00');
```
Fabricevladimir, line 47: Same thing here. I would use the formatter instead of a hardcoded value.

Fabricevladimir (commented):
Thanks, Kevin! Left a few comments and questsions

src/modules/utils/blockExplorerUtils/blockExplorerUtils.ts
```diff
@@ -0,0 +1,39 @@
...
+    ADDRESS = 'ADDRESS',
+}
+
+type ExplorerLinkCriteria = {
+    chainId?: number;
+    queryType?: QueryType;
+    txHash?: Hash;
+};
+
+export function createBlockExplorerLink(criteria: ExplorerLinkCriteria): string | undefined {
```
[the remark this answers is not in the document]
evanaronson: We do handle block explorer links poorly throughout the app, for example:
https://aragonassociation.atlassian.net/browse/APP-2616

Nevertheless, perhaps this is out of scope with this particular PR.

src/modules/utils/timestampUtils/timestampUtils.ts
```diff
@@ -0,0 +1,24 @@
+export const formatDate = (unixTimestamp?: number): string => {
```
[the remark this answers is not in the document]
thekidnamedkd: Should this just take a unix printout for now then as placeholder?

src/modules/components/transaction/transactionDataListItem/transactionDataListItemStructure/index.ts
```diff
@@ -0,0 +1,7 @@
+import { TransactionDataListItemStructure } from './transactionDataListItemStructure';
+
+export const TransactionDataListItem = {
+    Structure: TransactionDataListItemStructure,
+};
```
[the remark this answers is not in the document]
thekidnamedkd: That seems reasonable but this pattern matches the other DataListItem.Structures for the moment

src/modules/components/transaction/transactionDataListItem/transactionDataListItemStructure/transactionDataListItemStructure.api.ts
```diff
@@ -0,0 +1,71 @@
...
+     */
+    txStatus?: TxStatusCode;
+    /**
+     * The Unix timestamp of the transaction.
+     */
+    unixTimestamp?: number;
```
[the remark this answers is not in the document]
thekidnamedkd: Will leave as is for now?

src/modules/components/transaction/transactionDataListItem/transactionDataListItemStructure/transactionDataListItemStructure.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+    const getTxIcon = () => {
+        if (txStatus !== TxStatusCode.PENDING) {
+            return <AvatarIcon className="shrink-0" variant={variant} icon={icon} responsiveSize={{ md: 'md' }} />;
+        }
+        return (
+            <div className="flex size-6 shrink-0 items-center justify-center md:size-8">
+                <Spinner className="transition" variant="neutral" responsiveSize={{ md: 'lg' }} />
+            </div>
+        );
```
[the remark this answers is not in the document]
thekidnamedkd: I find this less readable personally, but it does match repo convention.

src/modules/components/transaction/transactionDataListItem/transactionDataListItemStructure/transactionDataListItemStructure.tsx
thekidnamedkd: Because Failed is ostensibly a TransactionType printout as the result of the status, but semantically it is not a type of transaction.

src/modules/components/transaction/transactionDataListItem/transactionDataListItemStructure/transactionDataListItemStructure.tsx
thekidnamedkd: We don't have variants for Heading at the moment. What would you suggest?

src/modules/components/transaction/transactionDataListItem/transactionDataListItemStructure/index.ts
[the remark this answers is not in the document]
Fabricevladimir: If the skeleton is in it's own folder, then the exports _have_ to be in the index of the parent because it won't make sense to be exporting the skeleton inside the index of the structure. The other DataListItem.Structure exports are incorrectly exported as well. Would you mind updating them so that they match the pattern set for the `ProposalDataListItem`?

src/modules/components/transaction/transactionDataListItem/transactionDataListItemStructure/transactionDataListItemStructure.tsx
thekidnamedkd: Yes, the Spinner has had updates which have not been implemented in the code, but represented in the design.

src/modules/utils/timestampUtils/timestampUtils.ts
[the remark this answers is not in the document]
Fabricevladimir: I would remove it altogether, allow the user to pass in a string (temporarily), and add a todo where the date is being used. See the `ProposalDataListItemStatus` component to see what I did in there. I understand it does require more work when the formatter is introduced since we have to make changes both in the ODS and the App, but the counter to that is the flexibility. Also, we need to figure out whether it makes sense to have the date formatter in the core vs the modules

src/modules/components/transaction/transactionDataListItem/transactionDataListItemStructure/transactionDataListItemStructure.tsx
Fabricevladimir: Aha, I see. Makes sense. I would revisit the function name though 🙈

src/modules/components/transaction/transactionDataListItem/transactionDataListItemStructure/transactionDataListItemStructure.tsx
Fabricevladimir: I left a question about whether these should be `Headings` in the first place. I wasn't able to find where that was indicated in the designs

src/modules/utils/blockExplorerUtils/blockExplorerUtils.ts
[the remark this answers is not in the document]
cgero-eth: `viem` / `wagmi` already give us the block explorer URLs. Given that we require the modules to be wrapped with a wagmi-provider, we can just take the urls from the provided chains. See [AddressInput](https://github.com/aragon/ods/blob/main/src/modules/components/address/addressInput/addressInput.tsx#L65) component

src/modules/components/transaction/transactionDataListItem/transactionDataListItemStructure/transactionDataListItemStructure.tsx
thekidnamedkd: They are not, but it seems logical to me?

src/modules/components/transaction/transactionDataListItem/transactionDataListItemStructure/transactionDataListItemStructure.api.ts
```diff
@@ -0,0 +1,71 @@
+import { type Hash } from 'viem';
+import { IconType, type AvatarIconVariant, type IDataListItemProps } from '../../../../../core';
+
+export enum TxStatusCode {
```
cgero-eth, line 4: To follow the naming of `TransactionType`, I would call this `TransactionStatus`
```suggestion
export enum TransactionStatus {
```

src/modules/components/transaction/transactionDataListItem/transactionDataListItemStructure/transactionDataListItemStructure.api.ts
```diff
@@ -0,0 +1,71 @@
...
+     */
+    tokenAddress?: string;
+    /**
+     * The symbol of the token, e.g. 'ETH' as a string
+     */
+    tokenSymbol?: string;
+    /**
+     * The token value in the transaction.
+     */
+    tokenValue?: number;
```
cgero-eth, line 50: I believe this is the amount of the tokens on the transaction, right? If so, I would probably call it `tokenAmount`

src/modules/components/transaction/transactionDataListItem/transactionDataListItemStructure/transactionDataListItemStructure.api.ts
```diff
@@ -0,0 +1,71 @@
...
+    /**
+     * The type of transaction.
+     */
+    txType: TransactionType;
+    /**
+     * The current status of a blockchain transaction on the network.
+     */
+    txStatus: TxStatusCode;
```
cgero-eth, line 58: Given that the component is called transaction-*, we don't need to add a `tx` prefix on props (as we do on other data-list components). We can rename these as `type` and `status`
```suggestion
    type: TransactionType;
    /**
     * The current status of a blockchain transaction on the network.
     */
    status: TxStatusCode;
```

src/modules/components/transaction/transactionDataListItem/transactionDataListItemStructure/transactionDataListItemStructure.api.ts
```diff
@@ -0,0 +1,71 @@
...
+    /**
+     * The Unix timestamp of the transaction.
+     */
+    formattedTimestamp?: string;
+    /**
+     * The estimated fiat value of the transaction.
+     */
+    fiatEstimate?: number;
```
cgero-eth, line 66: On the `assetDataListItem` component we ended up calculating the transaction value based on the token-price and token-amount. I would follow the same approach here, by adding a `tokenPrice` property and calculate the fiatAmount as `tokenAmount * tokenPrice`. 
@sepehr2github Handled these two properties to accept both `number | string`. I think it would be nice to do this here as well as sometimes we might get these data as strings or number

src/modules/components/transaction/transactionDataListItem/transactionDataListItemStructure/transactionDataListItemStructure.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+export const TransactionDataListItemStructure: React.FC<ITransactionDataListItemProps> = (props) => {
+    const {
+        chainId,
+        tokenAddress,
+        tokenSymbol,
+        tokenValue,
+        fiatEstimate,
+        txType = TransactionType.ACTION,
+        txStatus = TxStatusCode.PENDING,
```
cgero-eth, line 30: These are set as required on the prop interface, if we want to set default values we should set them as optional and specify the default values on the interface

src/modules/components/transaction/transactionDataListItem/transactionDataListItemStructure/transactionDataListItemStructure.tsx
```diff
@@ -0,0 +1,116 @@
...
+        blockExplorerBaseUrl && txHash ? `${blockExplorerBaseUrl}/tx/${txHash}` : undefined;
+
+    const parsedHref = blockExplorerAssembledHref ?? href;
+
+    const getEffectiveTxType = () => {
+        const type = txType;
+        if (txStatus === TxStatusCode.FAILED) {
+            return {
+                icon: IconType.CLOSE,
+                variant: 'critical' as AvatarIconVariant,
```
cgero-eth, line 53:
```suggestion
                variant: 'critical' as const,
```

src/modules/components/transaction/transactionDataListItem/transactionDataListItemStructure/transactionDataListItemStructure.tsx
```diff
@@ -0,0 +1,116 @@
...
+                            ? '-'
+                            : `${formattedTokenValue} ${tokenSymbol}`}
+                    </Heading>
+                    <Heading className="!text-neutral-500" size="h5" as="h2">
+                        {formatterUtils.formatNumber(
+                            fiatEstimate && txType !== TransactionType.ACTION ? fiatEstimate : 0,
+                            {
+                                format: NumberFormat.FIAT_TOTAL_SHORT,
+                            },
+                        )}
```
cgero-eth, line 110: For better readability, I would move these formatter calls before the return function as:
```
const formattedFiatAmount = formatterUtils.formatNumber(...)
```
And then simply:
```
<p>{formattedFiatAmount}</p>
```

src/modules/components/transaction/transactionDataListItem/transactionDataListItemStructure/transactionDataListItemStructure.tsx
cgero-eth: I believe these shouldn't be headings

src/modules/components/transaction/transactionDataListItem/transactionDataListItemStructure/transactionDataListItem.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+        {
+            id: '1',
+            blockExplorers: { default: { url: 'https://example.com' } },
+        },
+    ],
+}));
+
+describe('<TransactionDataListItemStructure /> component', () => {
+    const createTestComponent = (props?: Partial<ITransactionDataListItemProps>) => {
+        const defaultProps: ITransactionDataListItemProps = {
```
cgero-eth, line 21: The `defaultProps` should be set to handle only the required parameters

src/modules/components/transaction/transactionDataListItem/transactionDataListItemStructure/transactionDataListItem.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+jest.mock('wagmi', () => ({
+    useChains: () => [
+        {
+            id: '1',
+            blockExplorers: { default: { url: 'https://example.com' } },
+        },
+    ],
+}));
+
+describe('<TransactionDataListItemStructure /> component', () => {
```
cgero-eth, line 19:
```suggestion
describe('<TransactionDataListItem.Structure /> component', () => {
```

cgero-eth (approved):
LGTM, thanks @thekidnamedkd! ✅

PR #134, opened 2024-04-02
feat: APP-2977 - Implement AssetTransfer module component

thekidnamedkd:
## Description

implement AssetTransfer module component with tests and stories

Task: [APP-2977](https://aragonassociation.atlassian.net/browse/APP-2977)

## Type of change

## Checklist:

        the latest version.

[APP-2977]: https://aragonassociation.atlassian.net/browse/APP-2977?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/components/asset/assetTransfer/assetTransfer.stories.tsx
+++ b/src/modules/components/asset/assetTransfer/assetTransfer.stories.tsx
@@ -21,16 +21,18 @@ type Story = StoryObj<typeof AssetTransfer>;
 export const Default: Story = {
     args: {
         sender: { address: '0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045', name: 'vitalik.eth' },
-        recipient: { address: '0x168dAa4529bf88369ac8c1ABA5A2ad8CF2A61Fb9', name: 'decentralizedtransactions.eth' },
+        recipient: {
+            address: '0x168dAa4529bf88369ac8c1ABA5A2ad8CF2A61Fb9',
+            name: 'decentralizedtransactions.eth',
+        },
         assetIconSrc: 'https://assets.coingecko.com/coins/images/279/large/ethereum.png?1696501628',
         assetSymbol: 'ETH',
         assetAmount: 1,
         assetName: 'Ethereum',
         hash: '0xf006e9454ad77c5e8e6f54106c6939d3d8b68ae16fc216d67c752f54adb21fc6',
-        fiatPrice: 3850,
+        assetFiatPrice: 3850,
         chainId: 1,
     },
-    render: (props) => <AssetTransfer {...props} />,
 };
 
 /**
@@ -44,9 +46,7 @@ export const Fallback: Story = {
         assetSymbol: 'ETH',
         assetAmount: 1,
         hash: '0xf006e9454ad77c5e8e6f54106c6939d3d8b68ae16fc216d67c752f54adb21fc6',
-        chainId: 1,
     },
-    render: (props) => <AssetTransfer {...props} />,
 };
 
 export default meta;
--- a/src/modules/components/asset/assetTransfer/assetTransfer.stories.tsx
+++ b/src/modules/components/asset/assetTransfer/assetTransfer.stories.tsx
@@ -1,0 +1,36 @@
+import type { Meta, StoryObj } from '@storybook/react';
+import { AssetTransfer } from './assetTransfer';
+
+const meta: Meta<typeof AssetTransfer> = {
+    title: 'Modules/Components/Asset/AssetTransfer',
+    component: AssetTransfer,
+    tags: ['autodocs'],
+    parameters: {
+        design: {
+            type: 'figma',
+            url: 'https://www.figma.com/file/ISSDryshtEpB7SUSdNqAcw/branch/P0GeJKqILL7UXvaqu5Jj7V/Aragon-ODS?type=design&node-id=14385%3A24287&mode=dev&t=IX3Fa96hiwUEtcoA-1',
+        },
+    },
+};
+
+type Story = StoryObj<typeof AssetTransfer>;
+
+export const Default: Story = {};
+
+export const Loaded: Story = {
+    args: {
+        senderAddress: '0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045',
+        recipientAddress: '0x1D03D98c0aac1f83860cec5156116FE68725642E',
+        senderEnsName: 'vitalik.eth',
+        tokenIconSrc: 'https://assets.coingecko.com/coins/images/279/large/ethereum.png?1696501628',
+        tokenSymbol: 'ETH',
+        tokenAmount: 1,
+        tokenName: 'Ethereum',
+        hash: '0xf006e9454ad77c5e8e6f54106c6939d3d8b68ae16fc216d67c752f54adb21fc6',
+        tokenPrice: 3850,
+        chainId: 1,
+    },
+    render: (props) => <AssetTransfer {...props} />,
+};
+
+export default meta;
--- a/src/modules/components/asset/assetTransfer/assetTransfer.stories.tsx
+++ b/src/modules/components/asset/assetTransfer/assetTransfer.stories.tsx
@@ -15,19 +15,37 @@ const meta: Meta<typeof AssetTransfer> = {
 
 type Story = StoryObj<typeof AssetTransfer>;
 
-export const Default: Story = {};
-
-export const Loaded: Story = {
+/**
+ * Default usage example of the AssetTransfer component.
+ */
+export const Default: Story = {
     args: {
         senderAddress: '0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045',
-        recipientAddress: '0x1D03D98c0aac1f83860cec5156116FE68725642E',
+        recipientAddress: '0x168dAa4529bf88369ac8c1ABA5A2ad8CF2A61Fb9',
         senderEnsName: 'vitalik.eth',
+        recipientEnsName: 'decentralizedtransactions.eth',
         tokenIconSrc: 'https://assets.coingecko.com/coins/images/279/large/ethereum.png?1696501628',
-        tokenSymbol: 'ETH',
-        tokenAmount: 1,
+        symbol: 'ETH',
+        amount: 1,
         tokenName: 'Ethereum',
         hash: '0xf006e9454ad77c5e8e6f54106c6939d3d8b68ae16fc216d67c752f54adb21fc6',
-        tokenPrice: 3850,
+        fiatPrice: 3850,
+        chainId: 1,
+    },
+    render: (props) => <AssetTransfer {...props} />,
+};
+
+/**
+ * Fallback usage example of the AssetTransfer component with only required props.
+ */
+export const Fallback: Story = {
+    args: {
+        senderAddress: '0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045',
+        recipientAddress: '0x168dAa4529bf88369ac8c1ABA5A2ad8CF2A61Fb9',
+        tokenName: 'Ethereum',
+        symbol: 'ETH',
+        amount: 1,
+        hash: '0xf006e9454ad77c5e8e6f54106c6939d3d8b68ae16fc216d67c752f54adb21fc6',
         chainId: 1,
     },
     render: (props) => <AssetTransfer {...props} />,
--- a/src/modules/components/asset/assetTransfer/assetTransfer.stories.tsx
+++ b/src/modules/components/asset/assetTransfer/assetTransfer.stories.tsx
@@ -20,14 +20,12 @@ type Story = StoryObj<typeof AssetTransfer>;
  */
 export const Default: Story = {
     args: {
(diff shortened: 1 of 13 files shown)


src/modules/components/asset/assetTransfer/assetTransfer.tsx
```diff
@@ -0,0 +1,187 @@
...
+                </a>
+                <div className="border-t-[1px] border-neutral-100 md:border-l-[1px]" />
+                <AvatarIcon
+                    icon={IconType.CHEVRON_DOWN}
+                    size="sm"
+                    className={classNames(
+                        'absolute left-4 top-1/2 -translate-y-1/2 bg-neutral-50 text-neutral-300', //base
+                        'md:left-1/2 md:-translate-x-1/2 md:-rotate-90', //responsive
+                    )}
+                />
```
cgero-eth, line 139: So much code that could be reused, please consider implementing an internal `AssetTransferAddress` component to render the sender/recipient of the transfer

src/modules/components/asset/assetTransfer/assetTransfer.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+    return (
+        <div className="flex h-full w-[320px] flex-col gap-y-2 md:w-[640px] md:gap-y-3">
+            <div className="relative flex h-full  flex-col rounded-xl border-[1px] border-neutral-100 md:flex-row">
+                <AssetTransferAddress
+                    txRole="sender"
+                    ensName={senderEnsName}
+                    address={senderAddress}
+                    link={createLink(senderAddress)}
+                    handle={resolveHandle(senderEnsName, senderAddress)}
```
cgero-eth, line 117: The logic related to the `AssetTransferAddress` component should be moved there, same for tests. I'd suggest to create a `/assetTransferAddress` subfolder and move the component and related tests there

src/modules/components/asset/assetTransfer/assetTransfer.tsx
```diff
@@ -0,0 +1,150 @@
...
+    });
+    const fiatValue = Number(amount) * Number(fiatPrice);
+    const formattedFiatValue = formatterUtils.formatNumber(fiatValue, {
+        format: NumberFormat.FIAT_TOTAL_SHORT,
+        fallback: ` `,
+    });
+    const formattedTokenAmount = `${formattedTokenValue} ${tokenSymbol}`;
+
+    return { formattedTokenAmount, formattedFiatValue };
+};
```
thekidnamedkd, line 69: could be moved to a utils file for export? reusable in TranactionDataListItem.Structure for example

src/modules/components/asset/assetTransfer/assetTransfer.tsx
```diff
@@ -0,0 +1,150 @@
...
+    } = props;
+
+    const wagmiConfigProvider = useConfig();
+
+    const wagmiConfig = wagmiConfigProps ?? wagmiConfigProvider;
+
+    const processedChainId = chainId ?? wagmiConfig.chains[0].id;
+
+    const currentChain = wagmiConfig.chains.find(({ id }) => id === processedChainId);
+    const blockExplorerUrl = currentChain?.blockExplorers?.default.url;
```
thekidnamedkd, line 94: I did with this with useChains() in the TransactionDataListItem.Structure. I think this has a better fallback and scoped optionality for external users. But useChains() is also essentially bound to the config context for supported chain info from what I understand.

src/modules/components/asset/assetTransfer/assetTransfer.tsx
```diff
@@ -0,0 +1,150 @@
...
+                <div className="border-t-[1px] border-neutral-100 md:border-l-[1px]" />
+                <AvatarIcon
+                    icon={IconType.CHEVRON_DOWN}
+                    size="sm"
+                    className={classNames(
+                        'absolute left-4 top-1/2 -translate-y-1/2 bg-neutral-50 text-neutral-300', //base
+                        'md:left-1/2 md:-translate-x-1/2 md:-rotate-90', //responsive
+                    )}
+                />
+                <AssetTransferAddress
```
cgero-eth, line 126: The asset-transfer-address and asset-transfer-amount components are missing the white background color. Also double check the hover and active states, they look wrong

cgero-eth (commented):
Thanks for the implementation @thekidnamedkd

src/modules/components/asset/assetTransfer/assetTransfer.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+        recipient: { address: '0x168dAa4529bf88369ac8c1ABA5A2ad8CF2A61Fb9', name: 'decentralizedtransactions.eth' },
+        assetIconSrc: 'https://assets.coingecko.com/coins/images/279/large/ethereum.png?1696501628',
+        assetSymbol: 'ETH',
+        assetAmount: 1,
+        assetName: 'Ethereum',
+        hash: '0xf006e9454ad77c5e8e6f54106c6939d3d8b68ae16fc216d67c752f54adb21fc6',
+        fiatPrice: 3850,
+        chainId: 1,
+    },
+    render: (props) => <AssetTransfer {...props} />,
```
cgero-eth, line 33: The `render` function is probably redundant, here and on the story below

src/modules/components/asset/assetTransfer/assetTransfer.tsx
```diff
@@ -0,0 +1,130 @@
...
+import { Avatar, AvatarIcon, IconType, NumberFormat, formatterUtils } from '../../../../core';
+import { type ICompositeAddress, type IWeb3ComponentProps } from '../../../types';
+import { AssetTransferAddress } from './assetTransferAddress';
+
+export interface IRequiredCompositeAddress extends ICompositeAddress {
+    /**
+     * Required address the transfer participants.
+     */
+    address: string;
+}
```
cgero-eth, line 12: I would probably update the base `ICompositeAddress` to set the `address` as required, we usually want to have the same handling on all components.
@Fabricevladimir Do you see an use case where we only have the ens name and not the address of the `publisher` of a proposal?

src/modules/components/asset/assetTransfer/assetTransfer.tsx
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * Address of the transaction sender.
+     */
+    sender: IRequiredCompositeAddress;
+    /**
+     * Address of the transaction recipient.
+     */
+    recipient: IRequiredCompositeAddress;
+    /**
+     * Name of the token transferred.
```
cgero-eth, line 24: Let's update the comments as well
```suggestion
     * Name of the asset transferred.
```

src/modules/components/asset/assetTransfer/assetTransfer.tsx
```diff
@@ -0,0 +1,130 @@
...
+     */
+    assetAmount: number | string;
+    /**
+     * Symbol of the token transferred. Example: ETH, DAI, etc.
+     */
+    assetSymbol: string;
+    /**
+     * Price per token in fiat.
+     */
+    fiatPrice?: number | string;
```
cgero-eth, line 42: To be aligned with all the other properties
```suggestion
    assetFiatPrice?: number | string;
```

src/modules/components/asset/assetTransfer/assetTransfer.tsx
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * Transaction hash.
+     */
+    hash: string;
+    /**
+     * Chain ID of the transaction.
+     */
+    chainId: number;
```
cgero-eth, line 50: The `chainId` property is already set on the `IWeb3ComponentProps` interface as optional. Let's keep this optional and use the first chain set on the wagmi config as fallback. Remember to update the tests to only pass the required properties

src/modules/components/asset/assetTransfer/assetTransferAddress/assetTransferAddress.test.tsx
```diff
@@ -0,0 +1,69 @@
+import '@testing-library/jest-dom';
+import { render, screen } from '@testing-library/react';
+import { OdsModulesProvider } from '../../../odsModulesProvider';
+import { AssetTransferAddress, type IAssetTransferAddressProps } from './assetTransferAddress';
+
+jest.mock('../../../member/memberAvatar', () => ({ MemberAvatar: () => <div data-testid="member-avatar-mock" /> }));
```
cgero-eth, line 6: By using the index file we can change the internal structure on the component without affecting eventual imports, let's always import/mock from the index files
```suggestion
jest.mock('../../../member', () => ({ MemberAvatar: () => <div data-testid="member-avatar-mock" /> }));
```

src/modules/components/asset/assetTransfer/assetTransferAddress/assetTransferAddress.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+    };
+
+    it('renders correctly as a sender', () => {
+        const txRole = 'sender' as const;
+        render(createTestComponent({ txRole }));
+
+        expect(screen.getByText('From')).toBeInTheDocument();
+        // eslint-disable-next-line testing-library/no-node-access
+        const parentElement = screen.getByText('From').closest('a');
+        expect(parentElement).toHaveClass('rounded-t-xl md:rounded-l-xl md:rounded-r-none');
```
cgero-eth, line 29: As the component only has one link we could just use `getByRole('link')`, right? In general we should always try and use the `getByRole` and avoid disabling eslint rules

src/modules/components/asset/assetTransfer/assetTransferAddress/assetTransferAddress.tsx
```diff
@@ -0,0 +1,79 @@
...
+                className="shrink-0"
+                responsiveSize={{ md: 'md' }}
+                ensName={participant.name}
+                address={participant.address}
+            />
+            <div className="flex flex-col">
+                <span className="text-xs font-normal leading-tight text-neutral-500 md:text-sm">
+                    {txRole === 'sender' ? 'From' : 'To'}
+                </span>
+                <div className="grid grid-cols-[auto,1fr] items-center gap-x-1">
```
cgero-eth, line 70: We can probably make the text truncate by simply using flex and min-width: 0, it's a bit clearer than using a grid for a simple layout

cgero-eth (commented):
Component looks great now, well done @thekidnamedkd 👏🏻 
Left a couple of comments and a question for @Fabricevladimir

src/modules/components/asset/assetTransfer/assetTransferAddress/assetTransferAddress.tsx
thekidnamedkd: I was fussing with this, couldn't get the min-width: 0 to pick up before going to grid. I will try again.

src/modules/components/asset/assetTransfer/assetTransferAddress/assetTransferAddress.test.tsx
thekidnamedkd: Agreed @cgero-eth, but the issue is because we don't pass a blockExplorerLink in the test, as it's optional, then react-testing-library does not parse a 'link' role from the undefined fallback of the href. For this we have to go directly to the DOM to get the <a>.

src/modules/components/asset/assetTransfer/assetTransferAddress/assetTransferAddress.tsx
thekidnamedkd: Nice, got it. Need to go div higher to flex-col with the min-w-0.

src/modules/components/asset/assetTransfer/assetTransfer.tsx
cgero-eth: Sounds good, thanks @Fabricevladimir 👌🏻 
@thekidnamedkd Could you update the interface and update the proposal-list-item component to reflect such?

src/modules/components/asset/assetTransfer/assetTransferAddress/assetTransferAddress.test.tsx
cgero-eth: @thekidnamedkd I see, makes sense. As we need to access the link on every test, only for this component I would add the `blockExplorerLink` property to the completeProps object with a comment like:
```
const completeProps = {
    txRole: 'sender' as const,
    participant: { ... },
    // Set blockExplorerUrl parameter to retrieve the link element through the getByRole utility
    blockExplorerUrl: 'https://test.com',
```

src/modules/types/compositeAddress.ts
```diff
@@ -1,4 +1,4 @@
 export interface ICompositeAddress {
-    address?: string;
+    address: string;
```
cgero-eth, line 2: Could you update the `ProposalDataListItemStructure` component to reflect these changes?

cgero-eth (approved):
Thanks @thekidnamedkd!

PR #142, opened 2024-04-16
feat: APP-3054 - Implement StateSkeletonBar & StateSkeletonCircular core components

Fabricevladimir:
## Description

- implements the StateSkeletonBar and StateSkeletonCircular core components

Task: [APP-3054](https://aragonassociation.atlassian.net/browse/APP-3054)

## Type of change

## Checklist:

        the latest version.

[APP-3054]: https://aragonassociation.atlassian.net/browse/APP-3054?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/core/components/states/stateSkeletonBar/index.ts
+++ b/src/core/components/states/stateSkeletonBar/index.ts
@@ -1,5 +1,1 @@
-export {
-    StateSkeletonBar,
-    type IStateSkeletonBarProps,
-    type SkeletonBarSize as SkeletonSize,
-} from './stateSkeletonBar';
+export { StateSkeletonBar, type IStateSkeletonBarProps, type SkeletonBarSize } from './stateSkeletonBar';
--- a/src/core/components/states/stateSkeletonBar/index.ts
+++ b/src/core/components/states/stateSkeletonBar/index.ts
@@ -1,0 +1,5 @@
+export {
+    StateSkeletonBar,
+    type IStateSkeletonBarProps,
+    type SkeletonBarSize as SkeletonSize,
+} from './stateSkeletonBar';
--- a/src/core/components/states/stateSkeletonBar/index.ts
+++ b/src/core/components/states/stateSkeletonBar/index.ts
@@ -1,1 +1,1 @@
-export { StateSkeletonBar, type IStateSkeletonBarProps, type SkeletonBarSize } from './stateSkeletonBar';
+export { StateSkeletonBar, type IStateSkeletonBarProps, type StateSkeletonBarSize } from './stateSkeletonBar';
--- a/src/core/components/states/stateSkeletonBar/stateSkeletonBar.test.tsx
+++ b/src/core/components/states/stateSkeletonBar/stateSkeletonBar.test.tsx
@@ -1,0 +1,21 @@
+import { render, screen } from '@testing-library/react';
+import React from 'react';
+import { StateSkeletonBar } from './stateSkeletonBar';
+
+describe('<StateSkeletonBar /> component', () => {
+    const createTestComponent = (props?: Partial<React.ComponentPropsWithRef<typeof StateSkeletonBar>>) => {
+        const completeProps = { ...props };
+        return <StateSkeletonBar {...completeProps} />;
+    };
+
+    it('renders the skeleton element without crashing', () => {
+        render(createTestComponent());
+        expect(screen.getByTestId('stateSkeletonBar')).toBeInTheDocument();
+    });
+
+    it('forwards the ref to the skeleton element', () => {
+        const ref = React.createRef<HTMLDivElement>();
+        render(createTestComponent({ ref }));
+        expect(ref.current).toBe(screen.getByTestId('stateSkeletonBar'));
+    });
+});
--- a/src/core/components/states/stateSkeletonBar/stateSkeletonBar.test.tsx
+++ b/src/core/components/states/stateSkeletonBar/stateSkeletonBar.test.tsx
@@ -1,9 +1,8 @@
 import { render, screen } from '@testing-library/react';
-import React from 'react';
-import { StateSkeletonBar } from './stateSkeletonBar';
+import { StateSkeletonBar, type IStateSkeletonBarProps } from './stateSkeletonBar';
 
 describe('<StateSkeletonBar /> component', () => {
-    const createTestComponent = (props?: Partial<React.ComponentPropsWithRef<typeof StateSkeletonBar>>) => {
+    const createTestComponent = (props?: Partial<IStateSkeletonBarProps>) => {
         const completeProps = { ...props };
         return <StateSkeletonBar {...completeProps} />;
     };
@@ -11,11 +10,5 @@ describe('<StateSkeletonBar /> component', () => {
     it('renders the skeleton element without crashing', () => {
         render(createTestComponent());
         expect(screen.getByTestId('stateSkeletonBar')).toBeInTheDocument();
-    });
-
-    it('forwards the ref to the skeleton element', () => {
-        const ref = React.createRef<HTMLDivElement>();
-        render(createTestComponent({ ref }));
-        expect(ref.current).toBe(screen.getByTestId('stateSkeletonBar'));
     });
 });
--- a/src/core/components/states/stateSkeletonBar/stateSkeletonBar.tsx
+++ b/src/core/components/states/stateSkeletonBar/stateSkeletonBar.tsx
@@ -1,9 +1,15 @@
 import classNames from 'classnames';
 import { forwardRef, type ComponentPropsWithoutRef } from 'react';
+import { type ResponsiveAttribute, type ResponsiveAttributeClassMap } from '../../../types';
+import { responsiveUtils } from '../../../utils';
 
 export type SkeletonBarSize = 'sm' | 'md' | 'lg' | 'xl' | '2xl';
 
 export interface IStateSkeletonBarProps extends ComponentPropsWithoutRef<'div'> {
+    /**
+     *  Responsive size attribute for the skeleton.
+     */
+    responsiveSize?: ResponsiveAttribute<SkeletonBarSize>;
     /**
      * The size of the skeleton.
      * @default 'md'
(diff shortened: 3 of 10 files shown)


src/core/components/states/stateSkeletonBar/index.ts
```diff
@@ -0,0 +1,5 @@
+export {
+    StateSkeletonBar,
+    type IStateSkeletonBarProps,
+    type SkeletonBarSize as SkeletonSize,
```
cgero-eth, line 4: I'd suggest to use the component name as prefix and remove the alias to avoid confusions
```suggestion
    type StateSkeletonBarSize,
```

src/core/components/states/stateSkeletonBar/stateSkeletonBar.test.tsx
```diff
@@ -0,0 +1,21 @@
+import { render, screen } from '@testing-library/react';
+import React from 'react';
+import { StateSkeletonBar } from './stateSkeletonBar';
+
+describe('<StateSkeletonBar /> component', () => {
+    const createTestComponent = (props?: Partial<React.ComponentPropsWithRef<typeof StateSkeletonBar>>) => {
```
cgero-eth, line 6: Can't we just do:
```
Partial<IStateSkeletonBarProps>
```

src/core/components/states/stateSkeletonCircular/stateSkeletonCircular.tsx
```diff
@@ -0,0 +1,40 @@
...
+export const StateSkeletonCircular = forwardRef<HTMLDivElement, IStateSkeletonCircularProps>((props, ref) => {
+    const { className, responsiveSize = {}, size = 'md', ...otherProps } = props;
+
+    const classes = classNames(
+        'animate-pulse rounded-full bg-neutral-50',
+        responsiveUtils.generateClassNames(size, responsiveSize, responsiveSizeClasses),
+        className,
+    );
+
+    return <div data-testid="stateSkeletonCircular" ref={ref} className={classes} {...otherProps} />;
```
cgero-eth, line 37: Looks like there's a [radix-ui component](https://www.radix-ui.com/themes/docs/components/skeleton) for the skeleton loader 🙈 
Okay for me not to use it as it's a pretty basic component, but I would just check and apply the eventual accessibility roles

src/core/components/states/stateSkeletonBar/index.ts
Fabricevladimir: Oof, this is a vsCode rename. Good catch

src/core/components/states/stateSkeletonBar/stateSkeletonBar.test.tsx
Fabricevladimir: Not if we want to test for the forwarded ref unfortunately. That specific test isn't very useful though, at least not for this component. What do you think?

src/core/components/states/stateSkeletonCircular/stateSkeletonCircular.tsx
Fabricevladimir: Nah, it's not one of their primitives, it's in the themes. In terms of accessibility, I've been looking into this a bit, but it seems mostly hiding the component from the screen reader. The rest should be set on the parent element, in our case the DataListItem Skeleton, but I need to check whether we should put the aria attributes on the DataListItem Container instead considering we render multiple instances of the DataListItem Skeleton at once.

src/core/components/states/stateSkeletonBar/index.ts
cgero-eth: @Fabricevladimir wdyt about the renaming from `SkeletonBarSize` to `StateSkeletonBarSize`

src/core/components/states/stateSkeletonBar/stateSkeletonBar.test.tsx
cgero-eth: I think we can remove the `ref` test to be aligned with all the other components

src/core/components/states/stateSkeletonBar/index.ts
Fabricevladimir: It's coming 👍🏾

src/core/components/states/stateSkeletonCircular/stateSkeletonCircular.stories.tsx
```diff
@@ -0,0 +1,25 @@
...
+    },
+};
+
+type Story = StoryObj<typeof StateSkeletonCircular>;
+
+/**
+ * Default usage example of the StateSkeletonCircular component.
+ */
+export const Default: Story = {
+    args: {},
```
cgero-eth, line 22: Can we set a white background or a color with an higher contrast for these stories? It's hard to see the component with the current bg

src/core/components/states/stateSkeletonBar/stateSkeletonBar.tsx
```diff
@@ -0,0 +1,49 @@
...
+    lg: { sm: 'h-5', md: 'md:h-5', lg: 'lg:h-5', xl: 'xl:h-5', '2xl': '2xl:h-5' },
+    xl: { sm: 'h-6', md: 'md:h-6', lg: 'lg:h-6', xl: 'xl:h-6', '2xl': '2xl:h-6' },
+    '2xl': { sm: 'h-8', md: 'md:h-8', lg: 'lg:h-8', xl: 'xl:h-8', '2xl': '2xl:h-8' },
+};
+
+export const StateSkeletonBar = forwardRef<HTMLDivElement, IStateSkeletonBarProps>((props, ref) => {
+    const { className, responsiveSize = {}, size = 'md', ...otherProps } = props;
+
+    const classes = classNames(
+        'w-40 animate-pulse rounded-full bg-neutral-50',
```
cgero-eth, line 32: What's the plan for customising the width of the component?

Fabricevladimir: This is just a default width so the component does appear by default. When it's being used, we then override using classnames.

src/core/components/states/stateSkeletonCircular/stateSkeletonCircular.stories.tsx
Fabricevladimir: Defaulting to neutral-800

src/core/components/states/stateSkeletonBar/stateSkeletonBar.tsx
cgero-eth: Mm.. Have you tried overriding it on a Storybook story? I'm asking because we might need to always use the important modifier if we want to change the bar width which is not optimal. An alternative would be using a `width` attribute which set the width as inline style?

Fabricevladimir: Ah, right because no real idea in what order tailwind actually defines the actual CSS classes. `width` is a better approach since it doesn't make sense to have the component have zero width by default; not a huge fan because we essentially have 3 different ways to mess with the sizing of the component.

src/core/components/states/stateSkeletonBar/stateSkeletonBar.tsx
```diff
@@ -0,0 +1,62 @@
...
+     * Can be provided as a number (interpreted as pixels)
+     * or a string with explicit units (e.g., px, %, em).
+     * @default 160
+     *
+     * @example
+     * width={200}    // Sets width to 200 pixels.
+     * width="100px"  // Sets width to 100 pixels.
+     * width="10%"    // Sets width to 10% of its parent container.
+     */
+    width?: string | number;
```
cgero-eth, line 29:
```suggestion
    width?: CssProperties['width'];
```

src/core/components/states/stateSkeletonBar/stateSkeletonBar.tsx
```diff
@@ -0,0 +1,62 @@
...
+const responsiveSizeClasses: ResponsiveAttributeClassMap<StateSkeletonBarSize> = {
+    sm: { sm: 'h-3', md: 'md:h-3', lg: 'lg:h-3', xl: 'xl:h-3', '2xl': '2xl:h-3' },
+    md: { sm: 'h-4', md: 'md:h-4', lg: 'lg:h-4', xl: 'xl:h-4', '2xl': '2xl:h-4' },
+    lg: { sm: 'h-5', md: 'md:h-5', lg: 'lg:h-5', xl: 'xl:h-5', '2xl': '2xl:h-5' },
+    xl: { sm: 'h-6', md: 'md:h-6', lg: 'lg:h-6', xl: 'xl:h-6', '2xl': '2xl:h-6' },
+    '2xl': { sm: 'h-8', md: 'md:h-8', lg: 'lg:h-8', xl: 'xl:h-8', '2xl': '2xl:h-8' },
+};
+
+export const StateSkeletonBar = forwardRef<HTMLDivElement, IStateSkeletonBarProps>((props, ref) => {
+    const { className, responsiveSize = {}, size = 'md', width, ...otherProps } = props;
```
cgero-eth, line 41: To follow the `@default` directive, we should set the default value here `width = 160` and remove the `w-40` class

cgero-eth: We could also merge the style object with the style property to avoid eventual overrides:
`style={{ width, ...style }}`

Fabricevladimir: Yeah, I'm playing around with it to figure out responsiveness with the new inline style prop; looks like % are the way forward

src/core/components/states/stateSkeletonBar/stateSkeletonBar.tsx
Fabricevladimir: Pretty handy that 👍🏾

cgero-eth (approved):
LGMT, thanks @Fabricevladimir!

PR #147, opened 2024-04-19
feat: APP-3094 - Update ProposalDataListItemStrucure component

Fabricevladimir:
## Description

-   Add `stageId` and `stageTitle` properties to `IApprovalThresholdResult` & `IMajorityVotingResult` interfaces
-   Add `id` and optional `tag` properties to `ProposalDataListItemStructure`
-   Remove `publisherProfileLink` and `protocolUpdate` properties from `ProposalDataListItemStructure`
-   Update `date` and `result` properties of `ProposalDataListItemStructure` to be optional and `publisher` to allow for
    multiple publishers
-   Fixed `Link` core component children to truncate properly when overflowing

Task: [APP-3094](https://aragonassociation.atlassian.net/browse/APP-3094)

## Type of change

-   [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

        the latest version.

[APP-3094]: https://aragonassociation.atlassian.net/browse/APP-3094?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.api.ts
+++ b/src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.api.ts
@@ -24,9 +24,9 @@ export interface IProposalDataListItemStructureBaseProps<TType extends ProposalT
      */
     date?: string;
     /**
-     * Indicates whether the proposal is a protocol update
+     * Optional tag indicating proposal type
      */
-    protocolUpdate?: boolean;
+    tag?: string;
     /**
      * Publisher address (and optional ENS name)
      */
--- a/src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.api.ts
+++ b/src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.api.ts
@@ -60,24 +60,33 @@ export interface IProposalDataListItemStructureBaseProps<TType extends ProposalT
      */
     voted?: boolean;
 }
+export interface IPublisher extends ICompositeAddress {
+    /**
+     * Link to additional information about the publisher, such as a profile page or block explorer.
+     */
+    link?: string;
+}
 
 export interface IProposalStage {
     /**
      * Name of the proposal stage
      */
-    stageTitle?: string;
+    title?: string;
 
     /**
      * Id of the proposal stage
      */
-    stageId?: string | number;
+    id: string | number;
 }
 
-export interface IPublisher extends ICompositeAddress {
-    profileLink?: string;
+export interface IProposalResultBase {
+    /**
+     * Proposal stage
+     */
+    stage?: IProposalStage;
 }
 
-export interface IApprovalThresholdResult extends IProposalStage {
+export interface IApprovalThresholdResult extends IProposalResultBase {
     /**
      * Number of approvals for the proposal
      */
@@ -88,7 +97,7 @@ export interface IApprovalThresholdResult extends IProposalStage {
     approvalThreshold: number;
 }
 
-export interface IMajorityVotingResult extends IProposalStage {
+export interface IMajorityVotingResult extends IProposalResultBase {
     /**
      * Winning option
      */
--- a/src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.api.ts
+++ b/src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.api.ts
@@ -84,7 +84,7 @@ export interface IApprovalThresholdResult extends IProposalStage {
     approvalThreshold: number;
 }
 
-export interface IMajorityVotingResult {
+export interface IMajorityVotingResult extends IProposalStage {
     /**
      * Winning option
      */
--- a/src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.api.ts
+++ b/src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.api.ts
@@ -20,6 +20,10 @@ export interface IProposalDataListItemStructureBaseProps<TType extends ProposalT
     extends IDataListItemProps,
         IWeb3ComponentProps {
     /**
+     * Proposal id
+     */
+    id?: string;
+    /**
      * Indicates date relative to the proposal status
      */
     date?: string;
@@ -28,7 +32,7 @@ export interface IProposalDataListItemStructureBaseProps<TType extends ProposalT
      */
     tag?: string;
     /**
-     * Publisher address (and optional ENS name)
+     * Publisher(s) address (and optional ENS name and profile link)
      */
     publisher: IPublisher | IPublisher[];
     /**
@@ -66,7 +70,7 @@ export interface IProposalStage {
     /**
      * Id of the proposal stage
      */
-    stageId?: string;
+    stageId?: string | number;
 }
 
 export interface IPublisher extends ICompositeAddress {
--- a/src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.api.ts
+++ b/src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.api.ts
@@ -61,7 +61,19 @@ export interface IProposalDataListItemStructureBaseProps<TType extends ProposalT
     voted?: boolean;
 }
 
-export interface IApprovalThresholdResult {
+export interface IProposalStage {
+    /**
+     * Name of the proposal stage
+     */
+    stageTitle?: string;
+
+    /**
+     * Id of the proposal stage
+     */
+    stageId?: string;
+}
+
+export interface IApprovalThresholdResult extends IProposalStage {
     /**
      * Number of approvals for the proposal
      */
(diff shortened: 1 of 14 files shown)


src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.tsx
```diff
@@ -32,36 +32,61 @@ export const ProposalDataListItemStructure: React.FC<IProposalDataListItemStruct
...
-
-                {/* TODO: apply internationalization [APP-2627] */}
-                {protocolUpdate && <Tag label="OSx Update" variant="primary" />}
+                {tag && <Tag label={tag} variant="primary" />}
             </div>
         </DataList.Item>
     );
 };
+
+function parsePublisher(publisher: IPublisher, isConnected: boolean, connectedAddress: string | undefined) {
```
cgero-eth, line 85: I usually prefer defining function helpers before the component definition, it's like first you define the function and then you use it. Also here, just a personal preference :)

src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.tsx
```diff
@@ -32,36 +32,61 @@ export const ProposalDataListItemStructure: React.FC<IProposalDataListItemStruct
...
+    const publisherIsConnected = isConnected && connectedAddress?.toLowerCase() === publisher.address.toLowerCase();
```
cgero-eth, line 86: I'd suggest to always use the `isAddressEqual` functionality from viem to compare addresses

src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.api.ts
```diff
@@ -61,7 +61,23 @@ export interface IProposalDataListItemStructureBaseProps<TType extends ProposalT
...
+    stageTitle?: string;
+
+    /**
+     * Id of the proposal stage
+     */
+    stageId?: string | number;
+}
+
+export interface IPublisher extends ICompositeAddress {
+    profileLink?: string;
```
cgero-eth, line 77: We can call the property just `link`. I would also a short description for the property

src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.test.tsx
```diff
@@ -19,13 +20,12 @@ describe('<ProposalDataListItemStructure/> component', () => {
 
         const baseProps: Omit<IProposalDataListItemStructureProps, 'result'> = {
             date: new Date().toISOString(),
-            protocolUpdate: false,
-            publisher: { address: '0x123' },
+            tag: 'OSx Update',
+            publisher: { address: '0x123', profileLink: '#' },
```
cgero-eth, line 24: We can remove the `date` property as optional, right? Same for the `tag` and `profileLink` properties, let's only define the required properties here

src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.tsx
```diff
@@ -32,36 +32,61 @@ export const ProposalDataListItemStructure: React.FC<IProposalDataListItemStruct
...
 
-    const publisherIsConnected = isConnected && connectedAddress?.toLowerCase() === publisher.address.toLowerCase();
-    const publisherLabel = publisherIsConnected
-        ? 'You'
-        : publisher.name ?? addressUtils.truncateAddress(publisher.address);
+    const parsedPublisher = Array.isArray(publisher)
+        ? publisher.map((p) => parsePublisher(p, isConnected, connectedAddress))
+        : [parsePublisher(publisher, isConnected, connectedAddress)];
+
+    const showParsedPublisher = parsedPublisher.length <= 3;
```
cgero-eth, line 39: We can define a `export const maxProposalPublishers = 3` const above the component and use it for the tests as well if we want to change it later on

src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.test.tsx
```diff
@@ -158,9 +185,10 @@ describe('<ProposalDataListItemStructure/> component', () => {
             expect(screen.getByText(testProps.title)).toBeInTheDocument();
             expect(screen.getByText(testProps.summary)).toBeInTheDocument();
             expect(screen.getByText(testProps.status)).toBeInTheDocument();
-            expect(screen.getByText(testProps.date)).toBeInTheDocument();
+            expect(screen.getByText(testProps.date!)).toBeInTheDocument();
+            expect(screen.getByText(testProps.id!)).toBeInTheDocument();
             expect(
-                screen.getByText(addressUtils.truncateAddress(testProps.publisher.address ?? '')),
+                screen.getByText(addressUtils.truncateAddress((testProps.publisher as IPublisher).address)),
```
cgero-eth, line 191: Why is the cast to `IPublisher` needed here?

src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.test.tsx
```diff
@@ -82,16 +82,42 @@ describe('<ProposalDataListItemStructure/> component', () => {
...
             const testProps: IProposalDataListItemStructureProps = {
                 date: new Date().toISOString(),
-                publisher: { address: '0x123' },
-                publisherProfileLink: '#',
+                publisher: { address: '0x123', profileLink: '#' },
                 status: 'active',
                 summary: 'Example Summary',
                 title: 'Example Title',
                 type: 'approvalThreshold',
+                id: '0x1',
```
cgero-eth, line 120: Let move the testing of id and title properties outside the result type as not dependent by it. We could write it once and do not change both the result type tests every time we add, change, remove a property

cgero-eth (commented):
Thanks for the updates @Fabricevladimir!

src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.api.ts
```diff
@@ -20,25 +20,25 @@ export interface IProposalDataListItemStructureBaseProps<TType extends ProposalT
...
-     * Link to the publisher's profile
+     * Publisher(s) address (and optional ENS name and profile link)
      */
-    publisherProfileLink: string;
+    publisher: IPublisher | IPublisher[];
     /**
      * Result of the proposal shown only when it is active, challenged or vetoed.
      */
-    result: TType extends 'majorityVoting' ? IMajorityVotingResult : IApprovalThresholdResult;
+    result?: TType extends 'majorityVoting' ? IMajorityVotingResult : IApprovalThresholdResult;
```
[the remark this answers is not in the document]
Fabricevladimir: There is a case where we don't have results at all because the proposal is a draft.

src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.tsx
Fabricevladimir: In this case it overcomplicates things for no real benefit considering it does the same underlying calculation and more importantly, throws if either property isn't a valid `Address`. Both `address` and `connectedAddress` can be invalid addresses

src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.test.tsx
Fabricevladimir: It was needed since the `testProps` were typed, and `publisher` is `IPublisher | IPublisher[]`.

src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.tsx
cgero-eth: Oh I see, I think it would be great to have such an utility on `addressUtils` that we can reuse every time we need to compare addresses so that we don't forget to lower case it. It can be simply as:
```
const isAddressEqual = (addressOne?: string, addressTwo?: string) => this.isAddress(addressOne) && this.isAddress(addressTwo) && isAddressEqual(addressOne, addressTwo)
```
But I can create a ticket for it and do it on a separate PR

src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.test.tsx
cgero-eth: I see, makes sense 👍🏻

src/modules/components/proposal/proposalDataListItem/proposalDataListItemStructure/proposalDataListItemStructure.api.ts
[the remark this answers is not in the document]
cgero-eth: Okay makes sense 👍🏻

cgero-eth (approved):
Thanks for the updates @Fabricevladimir! ✅

PR #163, opened 2024-04-25
feat: APP-3057 - Implement DocumentParser core component

thekidnamedkd:
## Description

Implements wrapper component which can receive markdown or HTML docs as strings and sanitize the output while styling with Tailwind typography plugin.

Task: [APP-3057](https://aragonassociation.atlassian.net/browse/APP-3057)

## Type of change

## Checklist:

        the latest version.

[APP-3057]: https://aragonassociation.atlassian.net/browse/APP-3057?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


package.json
```diff
@@ -99,6 +103,7 @@
     "@testing-library/react": "^15.0.4",
     "@testing-library/user-event": "^14.5.2",
     "@types/jest": "^29.5.12",
+    "@types/sanitize-html": "^2",
```
cgero-eth, line 106: Let's specify the full version of the dependency

src/core/components/documentParser/documentParser.tsx
```diff
@@ -0,0 +1,56 @@
...
+import TipTapLink from '@tiptap/extension-link';
+import { EditorContent, useEditor } from '@tiptap/react';
+import { StarterKit } from '@tiptap/starter-kit';
+import classNames from 'classnames';
+import { useEffect, type ComponentProps } from 'react';
+import sanitizeHtml from 'sanitize-html';
+import { Markdown } from 'tiptap-markdown';
+
+export interface IDocumentParserProps extends Omit<ComponentProps<'div'>, 'ref'> {
+    stringDocument: string;
```
cgero-eth, line 11: Let's always document the component properties. I'd also call it just `document`, we don't usually prefix props with their types

cgero-eth (commented):
Thanks for the implementation @thekidnamedkd!

src/core/components/documentParser/documentParser.tsx
```diff
@@ -0,0 +1,56 @@
...
+                allowedClasses: {
+                    code: ['language-*', 'lang-*'],
+                },
+                allowedAttributes: {
+                    ...sanitizeHtml.defaults.allowedAttributes,
+                    img: ['src', 'alt'],
+                    a: ['href', 'title'],
+                },
+
+                disallowedTagsMode: 'recursiveEscape',
```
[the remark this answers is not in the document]
thekidnamedkd: https://www.npmjs.com/package/sanitize-html#default-options

we probably don't need allowedClasses, that is vestigial from trying to get syntax highlighting to work. but img and del (strikethrough) are not included by default:

https://www.npmjs.com/package/sanitize-html#default-options

and then disallowedTagsMode: 'recursiveEscape' is carryover from the current App implementation.

src/core/components/documentParser/documentParser.tsx
```diff
@@ -0,0 +1,56 @@
...
+            Image,
+            Markdown,
+            TipTapLink.configure({
+                openOnClick: false,
+            }),
+        ],
+    });
+
+    useEffect(() => {
+        if (parser) {
```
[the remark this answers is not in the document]
thekidnamedkd: useEditor hook is async initialization possibly returning null. This is also seen in the current App implementation.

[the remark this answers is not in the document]
cgero-eth: Here's how they initialise it on their docs:
```
const editor = useEditor({
    extensions,
    content,
  })
```
Having it set inside an useEffect is probably useful to update the content when the props changes, I'd still double check if this is needed or the `useEditor` hooks already handle that

src/core/components/documentParser/documentParser.tsx
```diff
@@ -0,0 +1,61 @@
+import { Image } from '@tiptap/extension-image';
+import TipTapLink from '@tiptap/extension-link';
+import { EditorContent, useEditor } from '@tiptap/react';
+import { StarterKit } from '@tiptap/starter-kit';
+import classNames from 'classnames';
+import { useEffect, type ComponentPropsWithoutRef } from 'react';
+import sanitizeHtml from 'sanitize-html';
+import { Markdown } from 'tiptap-markdown';
+
+export interface IDocumentParserProps extends Omit<ComponentPropsWithoutRef<'div'>, 'ref'> {
```
cgero-eth, line 10: By using `ComponentPropsWithoutRef` you don't need to Omit the ref property..

src/core/components/documentParser/documentParser.stories.tsx
```diff
@@ -0,0 +1,98 @@
...
+};
+
+type Story = StoryObj<typeof DocumentParser>;
+
+/**
+ * Default use case for the DocumentParser component with a basic string.
+ */
+export const Default: Story = {
+    args: {
+        stringDocument: 'Hello, world!',
```
cgero-eth, line 20: The stories needs to be updated to use the `document` property instead of `stringDocument`, right?

src/core/components/documentParser/documentParser.stories.tsx
```diff
@@ -0,0 +1,98 @@
...
+    },
+    render: (args) => <DocumentParser {...args} />,
```
cgero-eth, line 22: And the `render` function is not necessary here and on the stories below

cgero-eth (approved):
LGTM, thanks @thekidnamedkd!

PR #164, opened 2024-04-26
feat: APP-3114 - Support NextJs components

cgero-eth:
## Description

- Implement OdsCoreProvider, AvatarBase and LinkBase components
- Extend OdsModulesProvider to render the OdsCoreProvider as well

## Type of change

## Checklist:

        the latest version.


src/core/components/odsCoreProvider/odsCoreProvider.mdx
```diff
@@ -0,0 +1,17 @@
...
+import { OdsCoreProvider } from './odsCoreProvider';
+
+<Meta title="Core/Components/OdsCoreProvider" />
+
+<Title>OdsCoreProvider</Title>
+
+The `<OdsCoreProvider />` is a React context that defines the image and link elements to be rendered across all the ODS
+components, defaulting such values to `<img />` and `<a />` tags. The provider is used by the `<AvatarBase />` and
+`<LinkBase />` components, which are in turn used by all the other core and module components.
+
```
thekidnamedkd, line 11: Maybe we could add a bit more implementation detail here with including snippet for Next setup: 

```suggestion
import { OdsCoreProvider } from '@aragon-ods';
import Image from 'next/image';
import Link from 'next/link';

const App = () => {
  return (
    <OdsCoreProvider values={{ Link: Link, Img: Image }}>
      {/* The rest of your app components go here */}
    </OdsCoreProvider>
  );
};

export default App;
```

thekidnamedkd (commented):
Looking good! I think this is going to be very cool. Thanks for the implementation. Just a few questions around documentation for me.

src/core/components/odsCoreProvider/odsCoreProvider.mdx
cgero-eth: Yes makes sense 👍🏻

src/modules/components/odsModulesProvider/odsModulesProvider.tsx
```diff
@@ -31,18 +32,25 @@ export interface IOdsModulesProviderProps {
      * @default defaultQueryClient
      */
     queryClient?: QueryClient;
+    /**
+     * Values for the OdsCoreProvider context.
+     * @see IOdsCoreContext
+     */
+    coreProviderValues?: IOdsCoreProviderProps['values'];
```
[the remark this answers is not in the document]
cgero-eth: Yes, as you mentioned, the idea is that the `OdsModulesProvider` handle everything included on the `OdsCoreProvider` so that applications don't need to use both providers but only one. Even though the core-provider only handles the Link & Image components, I would still keep a generic name as core-values so that, if we add more properties to the core context, we don't need to update the OdsModulesProvider as well and change the property name

src/core/components/avatars/avatar/avatar.tsx
```diff
@@ -53,27 +54,31 @@ const responsiveSizeClasses: ResponsiveAttributeClassMap<AvatarSize> = {
 export const Avatar: React.FC<IAvatarProps> = (props) => {
     const { alt = 'avatar', className, fallback, responsiveSize = {}, size = 'sm', ...imageProps } = props;
 
+    const [imgLoading, setImgLoading] = useState(true);
+
     const containerClassNames = classNames(
-        'flex items-center justify-center overflow-hidden rounded-full',
+        'flex items-center justify-center overflow-hidden rounded-full [position:var(--ods-avatar-container-position)]',
```
thekidnamedkd, line 60: I think we have needed a `shrink-0` here fwiw. Almost all of our circle shapes do. Like Spinner and CircularSkeleton. They go oval in many use cases otherwise.

thekidnamedkd (approved):
Looks good! I'd say let's give it a shot.

src/core/components/avatars/avatar/avatar.tsx
cgero-eth: Yes makes sense, I'm adding it here on the base avatar component and removing it from the components using it

PR #197, opened 2024-05-28
feat: APP-2978 - Implement Wallet module component

thekidnamedkd:
## Description

Implement Wallet module component with tests & stories

Task: [APP-2798](https://aragonassociation.atlassian.net/browse/APP-2798)

## Type of change

## Checklist:

        the latest version.

[APP-2798]: https://aragonassociation.atlassian.net/browse/APP-2798?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/modules/components/wallet/wallet.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+export interface IWalletProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
+    /**
+     * Callback fired when the user connects their wallet.
+     */
+    onConnect?: () => void;
+    /**
+     * Callback fired when the user disconnects their wallet.
+     */
+    onDisconnect?: () => void;
```
cgero-eth, line 16: Given that we'll probably have a more complex handling for connecting / disconnecting the user on the app (display custom dialog for connect and display user dialog when user is connected), I wouldn't add here the logic of actually connecting and disconnecting the user.
I'm also not sure if we should couple this component to the connected user state on wagmi, I would probably just let the parent component pass the information about the connected user, we could use the `ICompositeAddress` interface to pass the user address and name.
This means that we can update the component properties to only add the `user: ICompositeAddress` property and trigger the onClick property whenever the user clicks on the wallet component

src/modules/components/wallet/wallet.tsx
```diff
@@ -0,0 +1,99 @@
...
+     */
+    connector?: 'injected' | 'walletConnect';
+    /**
+     * If selecting WalletConnect as the connector, you must pass WalletConnect project ID. Required for WalletConnect -- see https://docs.walletconnect.com/
+     */
+    projectId?: string;
+    /**
+     * If `true`, the wallet button will be disabled.
+     */
+    disabled?: boolean;
```
cgero-eth, line 28: The `disabled` property should already be defined on the extended button props

src/modules/components/wallet/wallet.tsx
```diff
@@ -0,0 +1,99 @@
...
+            {isDisconnected && <div>Connect</div>}
+
+            {connectPending && (
+                <>
+                    <div className="hidden md:mr-3 md:flex">
+                        <StateSkeletonBar className="shrink-0" width="64px" size="lg" />
+                    </div>
+                    <StateSkeletonCircular size="lg" />
+                </>
+            )}
```
cgero-eth, line 87: I believe that the loading state is supposed to be shown when we try to resolve the member avatar or the ens name, could you double check with Selim?

cgero-eth (commented):
Thanks for the implementation @thekidnamedkd!

src/modules/components/wallet/wallet.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+export interface IWalletProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
+    /**
+     * The connected user details.
+     */
+    user: ICompositeAddress;
+    /**
+     * Whether or not the app is connected to a wallet.
+     */
+    isConnected: boolean;
```
cgero-eth, line 16: If the user is not connected, we don't have an address to set. We can just remove the `isConnected` property, set the `user` property as optional and display the connect button / user avatar-ens depending if the `user` property is set

src/modules/components/wallet/wallet.tsx
```diff
@@ -0,0 +1,58 @@
...
+}
+
+export const Wallet: React.FC<IWalletProps> = (props) => {
+    const { user, isConnected = false, onClick, ...otherProps } = props;
+    const { data: ensName, isLoading: ensLoading } = useEnsName({
+        address: isConnected ? addressUtils.getChecksum(user.address) : undefined,
+        query: { enabled: isConnected },
```
cgero-eth, line 23: The query should only be enabled if the `user.name` property is not already set and if the `user` property is set

src/modules/components/wallet/wallet.tsx
```diff
@@ -0,0 +1,58 @@
...
+    });
+
+    const buttonClassName = classNames(
+        'flex h-12 items-center rounded-full border border-neutral-100 bg-neutral-0 px-4 py-1 text-neutral-500 transition-all',
+        'hover:border-neutral-200',
+        'focus:outline-none focus-visible:ring focus-visible:ring-primary focus-visible:ring-offset',
+        'active:bg-neutral-50 active:text-neutral-800',
+        {
+            '!p-1 md:!pl-4': isConnected,
+        },
```
cgero-eth, line 33: Let's avoid using the `!important` rule, it's usually a bad practice

src/modules/components/wallet/wallet.tsx
```diff
@@ -0,0 +1,58 @@
...
+}
+
+export const Wallet: React.FC<IWalletProps> = (props) => {
+    const { user, isConnected = false, onClick, ...otherProps } = props;
+    const { data: ensName, isLoading: ensLoading } = useEnsName({
+        address: isConnected ? addressUtils.getChecksum(user.address) : undefined,
+        query: { enabled: isConnected },
+    });
+
+    const buttonClassName = classNames(
```
cgero-eth, line 26: We also need to append an eventual `className` property

src/modules/components/wallet/wallet.tsx
```diff
@@ -0,0 +1,58 @@
...
+        'hover:border-neutral-200',
+        'focus:outline-none focus-visible:ring focus-visible:ring-primary focus-visible:ring-offset',
+        'active:bg-neutral-50 active:text-neutral-800',
+        {
+            '!p-1 md:!pl-4': isConnected,
+        },
+    );
+
+    return (
+        <button className={buttonClassName} onClick={onClick} {...otherProps}>
```
cgero-eth, line 37: No need to destructure the `onClick` property and pass it, just keep it on the `otherProps` object
```suggestion
        <button className={buttonClassName} {...otherProps}>
```

src/modules/components/wallet/wallet.tsx
```diff
@@ -0,0 +1,58 @@
...
+            {!isConnected && <div>Connect</div>}
```
cgero-eth, line 38: The `<div/>` tag is probably redundant
```suggestion
            {!isConnected && 'Connect'}
```

src/modules/components/wallet/wallet.tsx
```diff
@@ -0,0 +1,58 @@
...
+                    <div
+                        title={user.name ?? user.address}
+                        className="hidden min-w-0 max-w-24 items-center truncate md:mr-3 md:flex"
+                    >
+                        {user.name ??
+                            (ensLoading ? (
+                                <StateSkeletonBar size="lg" />
+                            ) : (
+                                ensName ?? addressUtils.truncateAddress(user.address)
+                            ))}
```
cgero-eth, line 51: `ensLoading` should only be set when actually loading the ENS, we can cleanup the checks here
```suggestion
                            ensLoading ? (
                                <StateSkeletonBar size="lg" />
                            ) : (
                                ensName ?? addressUtils.truncateAddress(user.address)
                            )
```

src/modules/components/wallet/wallet.stories.tsx
```diff
@@ -0,0 +1,55 @@
...
+ * Example usage of a connected Wallet component with only address.
+ */
+export const WithAddress: Story = {
+    args: {
+        user: {
+            address: '0x1234567890123456789012345678901234567890',
+        },
+        isConnected: true,
+    },
+    render: (props) => <Wallet {...props} />,
```
cgero-eth, line 38: Man, these redundant `render` function on the stories 👀

src/modules/components/wallet/wallet.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+        expect(button).toHaveTextContent('Connect');
+    });
+
+    it('renders member avatar when connected', () => {
+        const isConnected = true;
+        createTestComponent({ isConnected });
+        expect(screen.getByTestId('member-avatar-mock')).toBeInTheDocument();
+    });
+
+    it('renders user address when connected', () => {
```
cgero-eth, line 57:
```suggestion
    it('renders user address when connected and user has no ENS name linked', () => {
```

src/modules/components/wallet/wallet.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+    it('renders user name when provided and connected', () => {
+        const isConnected = true;
+        const user = {
+            address: '0x0987654321098765432109876543210987654321',
+            name: 'aragon.eth',
+        };
+        createTestComponent({ user, isConnected });
+        expect(screen.getByText('aragon.eth')).toBeInTheDocument();
+        expect(screen.queryByText('0x09…4321')).not.toBeInTheDocument();
+    });
```
cgero-eth, line 80: Would be nice to have these tests as well:
- renders user ENS when user address has a ENS name linked
- does not fetch user ENS when already provided on the user property

src/modules/components/wallet/wallet.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+export const Wallet: React.FC<IWalletProps> = (props) => {
+    const { user, isConnected = false, onClick, ...otherProps } = props;
+    const { data: ensName, isLoading: ensLoading } = useEnsName({
+        address: isConnected ? addressUtils.getChecksum(user.address) : undefined,
+        query: { enabled: isConnected },
+    });
+
+    const buttonClassName = classNames(
+        'flex h-12 items-center rounded-full border border-neutral-100 bg-neutral-0 px-4 py-1 text-neutral-500 transition-all',
```
cgero-eth, line 27: As we're already setting a vertical padding, the `h-12` shouldn't be needed

src/modules/components/wallet/wallet.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+    it('renders user address when connected and user has no ENS name linked', () => {
+        const user = {
+            address: '0x0987654321098765432109876543210987654321',
+        };
+        jest.spyOn(wagmi, 'useEnsName').mockReturnValue({
+            data: null,
+            isLoading: false,
+        } as wagmi.UseEnsNameReturnType);
```
cgero-eth, line 62:
```suggestion
        useEnsNameMock.mockReturnValue({
            data: null,
            isLoading: false,
        } as wagmi.UseEnsNameReturnType);
```

src/modules/components/wallet/wallet.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+        createTestComponent({ user });
+        expect(screen.getByText('0x09…4321')).toBeInTheDocument();
+    });
+
+    it('renders linked ENS name when connected and no user name provided', () => {
+        const user = {
+            address: '0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045',
+        };
+        jest.spyOn(wagmi, 'useEnsName').mockReturnValue({
```
cgero-eth, line 72:
```suggestion
        useEnsNameMock.mockReturnValue({
```

src/modules/components/wallet/wallet.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+        expect(screen.getByText('vitalik.eth')).toBeInTheDocument();
+        expect(screen.queryByText('0xd8…6045')).not.toBeInTheDocument();
+    });
+
+    it('renders user name provided when connected and does not fetch linked ENS name', () => {
+        const user = {
+            address: '0x0987654321098765432109876543210987654321',
+            name: 'vitalik.eth',
+        };
+        jest.spyOn(wagmi, 'useEnsName').mockReturnValue({
```
cgero-eth, line 87:
```suggestion
        useEnsNameMock.mockReturnValue({
```

cgero-eth (approved):
Still some minor cleanups on the wallet tests to be made but PR looks good, thanks @thekidnamedkd!

src/modules/components/wallet/wallet.tsx
cgero-eth: @thekidnamedkd It looks like this suggestion hasn't been applied?

thekidnamedkd: But don't we want the order to be "if available" user.name => ensName fetch => truncated address. This suggestion would remove provided user.name as first choice?

thekidnamedkd: Also for this reason I thought I understood your suggestion and simplified the outside check.

cgero-eth: You're right, the suggestion above does not consider the `user.name` attribute :/ In that case we could cleanup this a bit by doing something like:
```
const truncatedAddress
const displayedName = user?.name ?? ensName ?? addressUtils.truncateAddress(user?.address);

return (
  [...]
  {ensLoading ? <StateSkeletonBar size="lg" /> : displayedName}
```
Would this work?

thekidnamedkd: Perfect. This follows some similar patterns for `const resolvedXYZ = ...` elsewhere in ODS then :)

cgero-eth: 🙌🏻

src/modules/components/wallet/wallet.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+export const Default: Story = {};
+
+/**
+ * Example usage of a connected Wallet component with only address.
+ */
+export const WithAddress: Story = {
+    args: {
+        user: {
+            address: '0x1234567890123456789012345678901234567890',
+            name: 'TESTTTT',
```
cgero-eth, line 35:
```suggestion
```

cgero-eth (approved):

PR #217, opened 2024-06-20
fix(APP-3202): Ensure MemberAvatar component works on server environments

cgero-eth:
## Description

- Update MemberAvatar component to only call the blockies.create utility on client side
- Implement ssr-utils to check current environment
- Update MemberDataListItem component to accept votingpower set as string

Task: [APP-3202](https://aragonassociation.atlassian.net/browse/APP-3202)

## Type of change

## Checklist:

        the latest version.

[APP-3202]: https://aragonassociation.atlassian.net/browse/APP-3202?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/core/utils/ssrUtils/ssrUtils.ts
+++ b/src/core/utils/ssrUtils/ssrUtils.ts
@@ -1,0 +1,8 @@
+class SsrUtils {
+    /**
+     * Checks if code is running on a server environment.
+     */
+    isServer = () => typeof window === 'undefined';
+}
+
+export const ssrUtils = new SsrUtils();
--- a/src/modules/components/member/memberAvatar/memberAvatar.tsx
+++ b/src/modules/components/member/memberAvatar/memberAvatar.tsx
@@ -3,7 +3,7 @@ import type React from 'react';
 import { type Address } from 'viem';
 import { normalize } from 'viem/ens';
 import { useEnsAddress, useEnsAvatar, useEnsName } from 'wagmi';
-import { Avatar, type IAvatarProps } from '../../../../core';
+import { Avatar, ssrUtils, type IAvatarProps } from '../../../../core';
 import { addressUtils, ensUtils } from '../../../utils';
 
 export interface IMemberAvatarProps extends Omit<IAvatarProps, 'fallback'> {
@@ -45,9 +45,10 @@ export const MemberAvatar: React.FC<IMemberAvatarProps> = (props) => {
     });
     const resolvedAvatarSrc = avatarSrc ?? ensAvatarData ?? undefined;
 
-    const blockiesSrc = resolvedAddress
-        ? blockies.create({ seed: addressUtils.getChecksum(resolvedAddress), scale: 8, size: 8 }).toDataURL()
-        : undefined;
+    const blockiesSrc =
+        resolvedAddress && !ssrUtils.isServer()
+            ? blockies.create({ seed: addressUtils.getChecksum(resolvedAddress), scale: 8, size: 8 }).toDataURL()
+            : undefined;
 
     const isLoading = avatarLoading || nameLoading || addressLoading;
 
--- a/src/modules/components/member/memberDataListItem/memberDataListItemStructure/memberDataListItemStructure.tsx
+++ b/src/modules/components/member/memberDataListItem/memberDataListItemStructure/memberDataListItemStructure.tsx
@@ -15,7 +15,7 @@ export interface IMemberDataListItemProps extends IDataListItemProps {
     /**
      * The total voting power of the member.
      */
-    votingPower?: number;
+    votingPower?: number | string;
     /**
      * ENS name of the user.
      */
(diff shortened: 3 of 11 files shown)


src/modules/components/member/memberDataListItem/memberDataListItemStructure/memberDataListItemStructure.tsx
```diff
@@ -63,13 +69,13 @@ export const MemberDataListItemStructure: React.FC<IMemberDataListItemProps> = (
                     <div className="flex flex-col gap-y-2">
```
Fabricevladimir, line 69: - This here seems to be fixed sizing when `delegationCount` || `votingPower` is present.  This allows for the space between the `renderedHandle` and the `votingPower` as opposed to spacing growing under the `votingPower` when no delegation is present.

- Also, this might need to be discussed with design, but I think it might make sense to show votingPower based on whether it is undefined rather than when greater than zero. If not you have a very empty card hanging around.

image

Take a look at the delegations received designs on [figma](https://www.figma.com/design/ISSDryshtEpB7SUSdNqAcw/Aragon-ODS?node-id=17598-39947&m=dev)

cgero-eth: Agree, updated the component 👌🏻

src/modules/components/member/memberAvatar/memberAvatar.tsx
```diff
@@ -45,9 +45,10 @@ export const MemberAvatar: React.FC<IMemberAvatarProps> = (props) => {
     });
     const resolvedAvatarSrc = avatarSrc ?? ensAvatarData ?? undefined;
 
-    const blockiesSrc = resolvedAddress
-        ? blockies.create({ seed: addressUtils.getChecksum(resolvedAddress), scale: 8, size: 8 }).toDataURL()
-        : undefined;
+    const blockiesSrc =
+        resolvedAddress && !ssrUtils.isServer()
```
Fabricevladimir, line 49: Why are we forcing blockies to only render clientside?

src/core/utils/ssrUtils/ssrUtils.ts
```diff
@@ -0,0 +1,8 @@
+class SsrUtils {
+    /**
+     * Checks if code is running on a server environment.
+     */
+    isServer = () => typeof window === 'undefined';
```
Fabricevladimir, line 5: Very complicated utilities you have here...

src/modules/components/member/memberDataListItem/memberDataListItemStructure/memberDataListItemStructure.tsx
```diff
@@ -15,7 +16,7 @@ export interface IMemberDataListItemProps extends IDataListItemProps {
     /**
      * The total voting power of the member.
      */
-    votingPower?: number;
+    votingPower?: number | string;
```
Fabricevladimir, line 19: I expected string vp to be pre-formatted coming in, but then we are formatting it again..Would it make sense if it's a string to set the original string as the fallback value when formatting, since we've now no guarantee that the string can be parsed into a number? This might just be overcomplicating things though...

Fabricevladimir (approved):
Looking good!

src/core/utils/ssrUtils/ssrUtils.ts
cgero-eth: I love implementing such complex utils :D

src/modules/components/member/memberAvatar/memberAvatar.tsx
cgero-eth: Looks like `blockies` uses `document.createElement` to create a canvas and generate the address image => https://github.com/pedrouid/blockies-ts/blob/master/src/index.ts#L43
This is causing the new app-next repository to crash when we try to render the members avatars on the server side

src/modules/components/member/memberDataListItem/memberDataListItemStructure/memberDataListItemStructure.tsx
cgero-eth: Mm.. well we'll always get such parameters as strings from the backend (votingPower, balance, etc..), I believe we should keep handling the string types to avoid converting such strings to numbers on the apps

src/modules/components/member/memberAvatar/memberAvatar.tsx
Fabricevladimir: Ah, makes sense, thanks!

src/modules/components/member/memberDataListItem/memberDataListItemStructure/memberDataListItemStructure.tsx
Fabricevladimir: No getting it back as a string is fine. I'm thinking about the extensive external users assuming that this can be any string and not just a string that can be parsed into a number.

cgero-eth: Oh got it, yes it might be confusing for other devs. Mm.. maybe we can handle both formatted & unformatted strings in the future or just write a doc about such params

PR #224, opened 2024-06-25
feat(APP-2944): Implement utilities to format dates

cgero-eth:
## Description

- Implement utilities to format dates
- Add `globalSetup` file to set UTC timezone on tests

Task: [APP-2944](https://aragonassociation.atlassian.net/browse/APP-2944)

## Type of change

## Checklist:

        the latest version.

[APP-2944]: https://aragonassociation.atlassian.net/browse/APP-2944?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/core/utils/formatterUtils/formatterUtils.ts
```diff
@@ -94,9 +113,70 @@ class FormatterUtils {
         return formattedValue;
     };
 
-    private getDynamicOption = <TOptionValue extends string | number = number>(
-        value: number,
-        option?: DynamicOption<TOptionValue>,
+    formatDate = (value: DateTime | number | string | undefined, options: IFormatDateOptions = {}) => {
```
rcaragon, line 116: 🤔 Duration and Date are different entities, so it might be clearer to keep them separate. There might also be situations where we want to format the duration independently of the current time.
Then we would have two functions: `formatDate` and `formatDuration`.

rcaragon (dismissed):
Well done 👍

src/core/utils/formatterUtils/formatterUtils.ts
cgero-eth: Good point, as far as I know we currently only display durations based on the current date. In such cases, we still format a date as a duration. Given that this involves changes on DX and on Notion documentation, I would suggest to open an improvement ticket to further discuss internally if we should split this function it two and have sort of two formats like `NumberFormat` & `DurationFormat`, what do you think?

src/core/utils/formatterUtils/formatterUtils.mdx
```diff
@@ -114,6 +117,115 @@ of the total supply.
...
+    values={[
+        '2009-01-02T10:10:41.049Z',
+        '2023-10-23T15:33:12.049Z',
+        DateTime.utc().minus({ hours: 23 }).toISO(),
+        DateTime.utc().minus({ seconds: 1 }).toISO(),
+        DateTime.utc().plus({ hours: 25 }).toISO(),
+        DateTime.utc().plus({ months: 4 }).toISO(),
+    ]}
+    formats={[DateFormat.RELATIVE]}
+    isDate={true}
```
thekidnamedkd, line 230: I wonder if we should format as '9 months from now' instead of start with 'In 9 months' so that we keep the number at the top of the string

thekidnamedkd (approved):

thekidnamedkd (commented):
Looks good. I left a few comments and accidentally approved, but I think it's probably good to go besides maybe that one string conversation which isn't blocking.

src/core/utils/formatterUtils/formatterUtils.ts
rcaragon: Yes, sure, sounds good 👍

package.json
```diff
@@ -106,6 +107,7 @@
     "@testing-library/react": "^16.0.0",
     "@testing-library/user-event": "^14.5.2",
     "@types/jest": "^29.5.12",
+    "@types/luxon": "^3",
```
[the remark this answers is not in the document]
cgero-eth: Updated to set a fixed version 👌🏻

src/core/utils/formatterUtils/formatterUtils.mdx
cgero-eth: Good point, but I believe this is not supported by luxon :/

thekidnamedkd (approved):
Beautiful. Love it. Putting the Lux(on) in Luxury. 💎

PR #229, opened 2024-06-28
fix: update required params for useBlockExplorer hook, exports types for formatting dates

thekidnamedkd:
## Description

Make params for useBlockExplorer hook optional to handle DAO query requirements
use enum for more secure type enforcement of link "types" in construction
export DateFormat and other usages for DateFormatter util

## Type of change

## Checklist:

        the latest version.

### DIFF
--- a/src/modules/hooks/useBlockExplorer/useBlockExplorer.ts
+++ b/src/modules/hooks/useBlockExplorer/useBlockExplorer.ts
@@ -7,7 +7,7 @@ export enum ChainEntityType {
     TOKEN = 'token',
 }
 
-export interface IChainEntity {
+export interface IGetChainEntityUrlParams {
     /**
      * The type of the chain entity (address, tx, token)
      */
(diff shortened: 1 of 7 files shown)


src/modules/hooks/useBlockExplorer/useBlockExplorer.ts
```diff
@@ -1,7 +1,11 @@
 import { useCallback } from 'react';
 import { useChains, type Config } from 'wagmi';
 
-type ChainEntityType = 'address' | 'tx' | 'token';
+export enum ChainEntityType {
```
rcaragon, line 4: I'm curious, how enum would provide more secure type enforcement? I never thought about that before 🤔

rcaragon (dismissed):

src/modules/hooks/useBlockExplorer/useBlockExplorer.ts
cgero-eth: I believe it doesn't make a big difference here, how I see it is that here the enum provides a sort of abstraction of the underlying value. If the block explorer path for transactions changes from `tx` to `transaction` (which probably never happen in any case), we only need to change the enum value

cgero-eth (approved):

PR #234, opened 2024-07-09
fix(APP-2994): Remove min width from CardEmptyState and add alignment

shan8851:
## Description

- Remove fixed width from `EmptyState` component
- Center `CardEmptyState` component

Task: [APP-2994](https://aragonassociation.atlassian.net/browse/APP-2994)

## Type of change

-   [ ] New feature (non-breaking change which adds functionality)
-   [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

-   [ ] I have commented my code, particularly in hard-to-understand areas.
-   [ ] I have made corresponding changes to the documentation.
-   [ ] Any dependent changes have been merged and published in downstream modules.
        the latest version.

[APP-2994]: https://aragonassociation.atlassian.net/browse/APP-2994?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shan8851:
@cgero-eth I didn't update the changelog here, feels like a super minor change, but let me know if we should update it for everything or not.

src/core/components/states/emptyState/emptyState.tsx
```diff
@@ -14,7 +14,7 @@ export const EmptyState: React.FC<IEmptyStateProps> = ({
     humanIllustration,
 }) => {
     const containerClassNames = classNames(
-        'grid w-[320px] md:w-[640px]', // Defaults
+        'grid w-full max-w-sm', // Defaults
```
cgero-eth, line 17: Should we also update the fixed widths below? The IllustrationHuman component for instance has a ficed width set to `400px` and it will probably overflow if the component is placed on a container that does not have enough space

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+-   Remove min width from `CardEmptyState`
+-   Centralize `CardEmptyState` component
```
cgero-eth, line 11:
```suggestion
-   Remove fixed width from `EmptyState` core component
-   Center `CardEmptyState` core component
```

cgero-eth (commented):
Thanks for the fixes @shan8851!

src/core/components/states/emptyState/emptyState.tsx
shan8851: @cgero-eth thanks for this, so I agree having fixed widths is generally not great, but in this case the 400px is only for `md:` breakpoints, so I don't think it will overflow. For small devices it is 295 and given these are so specific I wasn't sure about changing them and breaking image ratio or something like that. What do you think?

cgero-eth: I would still remove these fixed widths and make sure ratio and UI still works fine. Even if 400px width is set only for md breakpoint, the component itself might be placed inside a small container on md breakpoint and still overflow

shan8851: Yeah makes sense. I removed the 400px, but left some others as the explicit widths helped on smaller devices as the human illustration has a large frame so using full width would actually make the image look quite small.

It now looks nice on all configurations and screen sizes

shan8851: 🔫

cgero-eth (dismissed):
Thanks @shan8851!

cgero-eth (approved):
LGTM, thanks @shan8851!

PR #244, opened 2024-07-16
feat(APP-3059): Implement ProposalActions module component

thekidnamedkd:
# Description

Implement central logic for ProposalActionsTerminal with WithdrawToken actions as first example. 

Tasks: 

[APP-3354](https://aragonassociation.atlassian.net/browse/APP-3354)
[APP-3356](https://aragonassociation.atlassian.net/browse/APP-3356)

## Type of Change

## Checklist:

-   [x] I have tested my code locally.
-   [x] My code follows the style guidelines of this project
-   [x] I have added tests that prove my fix is effective or that my feature works

[APP-3354]: https://aragonassociation.atlassian.net/browse/APP-3354?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[APP-3356]: https://aragonassociation.atlassian.net/browse/APP-3356?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


CHANGELOG.md
```diff
@@ -9,10 +9,16 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [1.0.39] - 2024-07-16
 
+### Added
```
cgero-eth, line 12: These entries should be under the `unreleased` header

src/modules/components/proposal/proposalActions/utils/proposalActionsUtils.ts
```diff
@@ -0,0 +1,19 @@
+import { ProposalActionWithdrawToken } from '../actions/proposalActionWithdrawToken/proposalActionWithdrawToken';
```
cgero-eth, line 1: As we won't have more utility files, I'd move the utils and test files to the `/proposalActions` folder directly

src/modules/components/proposal/proposalActions/proposalActionsTypes/proposalAction.ts
```diff
@@ -0,0 +1,60 @@
+export enum ProposalActionType {
+    WITHDRAW_TOKEN = 'withdrawToken',
+}
```
cgero-eth, line 3: I would move the enum to a separate file `ProposalActionType` file

src/modules/components/proposal/proposalActions/utils/proposalActionsUtils.ts
```diff
@@ -0,0 +1,19 @@
...
+import type { IProposalAction, IProposalActionWithdrawToken } from '../proposalActionsTypes';
+import { ProposalActionType } from '../proposalActionsTypes/proposalAction';
```
cgero-eth, line 3:
```suggestion
import { type IProposalAction, type IProposalActionWithdrawToken, ProposalActionType } from '../proposalActionsTypes';
```

src/modules/components/proposal/proposalActions/utils/proposalActionsUtils.ts
```diff
@@ -0,0 +1,19 @@
...
+class ProposalActionsUtils {
+    public getActionComponent(action: IProposalAction) {
+        if (this.isWithdrawTokenAction(action)) {
+            return () => ProposalActionWithdrawToken({ action });
+        }
+
+        return null;
+    }
+
+    public isWithdrawTokenAction(action: Partial<IProposalAction>): action is IProposalActionWithdrawToken {
```
cgero-eth, line 14: Just to follow how we're currently defining functions in utilities
```suggestion
   getActionComponent = (action: IProposalAction) => {
        if (this.isWithdrawTokenAction(action)) {
            return () => ProposalActionWithdrawToken({ action });
        }

        return null;
    }

    isWithdrawTokenAction = (action: Partial<IProposalAction>): action is IProposalActionWithdrawToken => {
```

src/modules/components/proposal/proposalActions/utils/proposalActionsUtils.ts
cgero-eth, line 1:
```suggestion
import { ProposalActionWithdrawToken } from '../actions/proposalActionWithdrawToken';
```

src/modules/components/proposal/proposalActions/utils/proposalActionsUtils.test.tsx
```diff
@@ -0,0 +1,25 @@
+import { render, screen } from '@testing-library/react';
+import { generateProposalActionGeneric } from '../actions/generators/proposalActionGeneric';
+import { generateProposalActionWithdrawToken } from '../actions/generators/proposalActionWithdrawToken';
+import { proposalActionsUtils } from './proposalActionsUtils';
+
+jest.mock('../actions/proposalActionWithdrawToken/proposalActionWithdrawToken', () => ({
+    ProposalActionWithdrawToken: jest.fn(() => <div>Mock ProposalActionWithdrawToken</div>),
```
cgero-eth, line 7:
```suggestion
    ProposalActionWithdrawToken: () => <div>Mock ProposalActionWithdrawToken</div>,
```

src/modules/components/proposal/proposalActions/proposalActionsTypes/proposalAction.ts
```diff
@@ -0,0 +1,60 @@
...
+     * The parameters to pass to the function
+     */
+    parameters: IProposalActionInputDataParameter[];
+}
+
+export interface IProposalAction {
+    /**
+     * The address of the contract the proposal is attempting to interact with
+     */
+    contractAddress: string;
```
cgero-eth, line 35: Isn't the `to` parameter already the `contractAddress`?

src/modules/components/proposal/proposalActions/proposalActionsContainer/proposalActionsContainer.tsx
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        if (isValidElement(children)) {
+            return cloneElement(children, {
+                onToggle: () => handleToggleItem(0),
+            });
+        }
+
+        return children;
+    };
```
cgero-eth, line 68: As we're setting a callback on the accordion.container, we don't probably need to pass a `onToggle` property to each child component. Radix-ui should trigger the callback set on the accordion-container when a child component open state changes

src/modules/components/proposal/proposalActions/proposalActionsContainer/proposalActionsContainer.tsx
```diff
@@ -0,0 +1,94 @@
...
+    const [expandedItems, setExpandedItems] = useState<string[]>([]);
+    const { copy } = useOdsModulesContext();
+
+    const childrenCount = Array.isArray(children) ? children.length : children ? 1 : 0;
+
+    const handleToggleAll = () => {
+        if (expandedItems.length === childrenCount) {
+            setExpandedItems([]);
+        } else {
+            setExpandedItems(Array.from({ length: childrenCount }, (_, index) => `${index}-action`));
```
cgero-eth, line 29: I'd set the index to the actions simply to `index` to avoid having to build the action id every time

src/modules/components/proposal/proposalActions/proposalActionsContainer/proposalActionsContainer.tsx
```diff
@@ -0,0 +1,94 @@
...
+     */
+    className?: string;
+}
+
+export const ProposalActionsContainer: React.FC<IProposalActionsContainerProps> = (props) => {
+    const { className, children } = props;
+    const [expandedItems, setExpandedItems] = useState<string[]>([]);
+    const { copy } = useOdsModulesContext();
+
+    const childrenCount = Array.isArray(children) ? children.length : children ? 1 : 0;
```
cgero-eth, line 23: `Children.toArray(children).length` should work the same

src/modules/components/proposal/proposalActions/proposalActionsContainer/proposalActionsContainer.tsx
```diff
@@ -0,0 +1,94 @@
...
+    };
+
+    return (
+        <Card className={classNames('w-full overflow-hidden', className)}>
+            <Heading size="h2" className="px-4 pt-4 md:px-6 md:pt-6">
+                {copy.proposalActionsContainer.containerName}
+            </Heading>
+            <Accordion.Container isMulti={true} value={expandedItems} onValueChange={handleAccordionValueChange}>
+                {renderChildren()}
+                <div className="mt-1 flex w-full flex-col gap-y-3 overflow-hidden px-4 pb-4 md:flex-row-reverse md:px-6 md:pb-6">
```
cgero-eth, line 77: Why is the overflow-hidden class needed here and on the div below?

src/modules/components/proposal/proposalActions/proposalActionsContainer/proposalActionsContainer.tsx
```diff
@@ -0,0 +1,94 @@
...
+                            {expandedItems.length === childrenCount
+                                ? copy.proposalActionsContainer.collapse
+                                : copy.proposalActionsContainer.expand}
+                        </Button>
+                    )}
+                    {/** TODO: button cluster logic for states and tx responses **/}
+                    {/** TODO: needs logic for clearing message as more buttons become available (execute, etc) **/}
+                    <p className="w-full text-center text-sm text-neutral-500 md:text-start">
+                        {copy.proposalActionsContainer.footerMessage}
+                    </p>
```
cgero-eth, line 89: Mm.. I probably overlook the discussion here. Given that we need to make this portion of the component customisable (by adding the execute buttons or alerts), we'll probably need to use the `children` property here.
Would it work if instead we go back to your previous implementation of passing the actions list to the `proposalActions.container` component and render the `children` property here instead?
Also the `footerMessage` should not be handled by this component but passed by the parent component

src/modules/components/proposal/proposalActions/proposalActionsContainer/proposalActionsContainer.tsx
```diff
@@ -0,0 +1,94 @@
+import classNames from 'classnames';
+import { cloneElement, isValidElement, useState, type ReactElement } from 'react';
+import { Accordion, Button, Card, Heading, type IAccordionContainerBaseProps } from '../../../../../core';
+import { useOdsModulesContext } from '../../../odsModulesProvider';
+import { type IProposalActionsActionProps } from '../proposalActionsAction';
+
+export interface IProposalActionsContainerProps extends Omit<IAccordionContainerBaseProps<true>, 'isMulti'> {
```
cgero-eth, line 7: Do we need to expose any accordion-container specific prop? Probably no for now, I'd simply extend the `ComponentProps<'div'>` interface and pass all other-props to the card component (make sure to avoid overriding the `className` component)

src/modules/components/proposal/proposalActions/actions/generators/proposalActionGeneric.ts
```diff
@@ -0,0 +1,12 @@
+import type { IProposalAction } from '../../proposalActionsTypes';
+
+export const generateProposalActionGeneric = (action?: Partial<IProposalAction>): IProposalAction => ({
```
cgero-eth, line 3: Usually these generators name matches the entity name, in this case I'd call this `generateProposalAction` and change the name as well

src/modules/components/proposal/proposalActions/actions/generators/proposalActionWithdrawToken.ts
```diff
@@ -0,0 +1,45 @@
+import { type ICompositeAddress } from '../../../../../types';
+import type { IProposalActionWithdrawToken } from '../../proposalActionsTypes';
+import { ProposalActionType } from '../../proposalActionsTypes/proposalAction';
+
+export const generateCompositeAddress = (address?: Partial<ICompositeAddress>): ICompositeAddress => ({
+    address: '0x0000000000000000000000000000000000000000',
+    name: '',
```
cgero-eth, line 7: The `name` on `compositeAddress` interface is optional
```suggestion
```

src/modules/components/proposal/proposalActions/actions/proposalActionWithdrawToken/proposalActionWithdrawToken.tsx
```diff
@@ -0,0 +1,25 @@
...
+    const { action } = props;
+    return (
+        <AssetTransfer
+            sender={action.sender}
+            recipient={action.receiver}
+            assetName={action.token.name}
+            assetAmount={action.amount}
+            assetSymbol={action.token.symbol}
+            assetIconSrc={action.token.logo}
+            // TODO: Amend AssetTransfer hash req for proposal usage where no Tx exists yet, should disable asset link out
```
cgero-eth, line 21: Could you create a Jira ticket and link it here?
```suggestion
            // TODO: Make hash property on AssetTransfer optional (APP-XXXX)
```

src/modules/components/proposal/proposalActions/actions/proposalActionWithdrawToken/proposalActionWithdrawToken.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+describe('<ProposalActionWithdrawToken /> component', () => {
+    const createTestComponent = (props?: Partial<IProposalActionWithdrawToken>) => {
+        const defaultProps: IProposalActionWithdrawTokenProps = {
+            action: generateProposalActionWithdrawToken(props),
+        };
+
+        return <ProposalActionWithdrawToken {...defaultProps} />;
+    };
+
+    it('renders without crashing', () => {
```
cgero-eth, line 20: The `renders without crashing` name is very generic, here you're testing that it renders the AssetTransfer component
```suggestion
    it('renders the AssetTransfer component', () => {
```

src/modules/components/proposal/proposalActions/index.ts
```diff
@@ -0,0 +1,10 @@
+import { ProposalActionsAction } from './proposalActionsAction';
+import { ProposalActionsContainer } from './proposalActionsContainer';
+
+export const ProposalActions = {
+    Container: ProposalActionsContainer,
+    Action: ProposalActionsAction,
+};
+
+export * from './actions';
```
cgero-eth, line 9: We usually also export everything defined in the sub-components:
export * from './proposalActionsAction';
export * from './proposalActionsContainer';

src/modules/components/proposal/proposalActions/proposalActionsTypes/proposalAction.ts
thekidnamedkd: Not sure I follow exactly, but thinking a bit more about it if so, I guess it would be the 'from' then maybe? As in we are proposing that a treasury contract sends money 'from' itself 'to' a recipient contract/address??

src/modules/components/proposal/proposalActions/proposalActionsAction/proposalActionsAction.tsx
```diff
@@ -0,0 +1,54 @@
...
+    const ActionComponent = proposalActionsUtils.getActionComponent(action);
+    const { copy } = useOdsModulesContext();
+
+    const actionTypeToStringMapping: Record<string, string> = {
+        withdrawToken: copy.proposalActionsAction.actionTypeWithdrawToken,
+    };
+
+    if (!ActionComponent) {
+        return null;
+    }
```
cgero-eth, line 33: We should still render something when we don't have a specific UI for an action type

src/modules/components/proposal/proposalActions/proposalActionsAction/proposalActionsAction.tsx
```diff
@@ -0,0 +1,54 @@
...
+    index: number;
+    /**
+     * Toggle accordion callback for notifying the parent container to update after click
+     */
+    onToggle?: () => void;
+}
+
+export const ProposalActionsAction: React.FC<IProposalActionsActionProps> = (props) => {
+    const { action, index, onToggle } = props;
+    const ActionComponent = proposalActionsUtils.getActionComponent(action);
```
cgero-eth, line 24: This spec on Jira is currently missing:
> - We also want to allow users to render their own components for actions not handled by the ODS library, users should also be able to render their own UI for actions handled by the ODS library. In order to do so, we could update the  ProposalActionsContainer component to handle a "getActionComponent" property.

src/modules/components/proposal/proposalActions/proposalActionsContainer/proposalActionsContainer.tsx
thekidnamedkd: <div> child is overflowing the card and adding right angles to the bottom, class needs to be added to `Card` itself. Will adjust and log.

src/modules/components/proposal/proposalActions/proposalActionsAction/proposalActionsAction.tsx
```diff
@@ -0,0 +1,54 @@
...
+    const { copy } = useOdsModulesContext();
+
+    const actionTypeToStringMapping: Record<string, string> = {
+        withdrawToken: copy.proposalActionsAction.actionTypeWithdrawToken,
+    };
```
cgero-eth, line 29: Given that we also want to make the custom-actions customisable, it won't be possible to customise the action names in that way. What if we instead add an `actionNames` optional property (`Record<string, string>`)  to the container component and fallback to the function name when available?
It can be something like:
```
const actionNames = Object.keys(ActionTypes).reduce((accumulator, current) => ({ ...accumulator, [current]: t(`app.governance.action.${current}`), {});

<Action.Container actions={actions} actionNames={actionNames} />
```

The property can be then passed to the action component (if having the actions property set on the container) as:
<Action.Action action={action} name={actionNames?.[action.type]} />

src/modules/components/proposal/proposalActions/actions/proposalActionWithdrawToken/proposalActionWithdrawToken.tsx
thekidnamedkd: Will do! We were requiring it before because it would break the link creator without it, but can be handled.

src/modules/components/proposal/proposalActions/proposalActionsActionVerfication/proposalActionsActionVerfication.tsx
```diff
@@ -0,0 +1,56 @@
+import { type HTMLAttributes } from 'react';
+import { Heading, Icon, IconType } from '../../../../../core';
+import { addressUtils } from '../../../../utils';
+import type { IProposalAction, IProposalActionWithdrawToken } from '../proposalActionsTypes';
+import { ProposalActionType } from '../proposalActionsTypes/proposalAction';
```
cgero-eth, line 5:
```suggestion
import { type IProposalAction, type IProposalActionWithdrawToken, ProposalActionType } from '../proposalActionsTypes';
```

src/modules/components/proposal/proposalActions/proposalActionsActionVerfication/proposalActionsActionVerfication.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+export interface IProposalActionsActionVerificationProps extends HTMLAttributes<HTMLDivElement> {
```
cgero-eth, line 7:
```suggestion
export interface IProposalActionsActionVerificationProps extends ComponentProps<'div'> {
```

src/modules/components/proposal/proposalActions/proposalActionsAction/proposalActionsAction.tsx
thekidnamedkd: Should I add a TODO and await a design?

src/modules/components/proposal/proposalActions/proposalActionsActionVerfication/proposalActionsActionVerfication.tsx
```diff
@@ -0,0 +1,56 @@
...
+                verificationLabel = withdrawAction.token.name;
+                break;
+            }
+
+            default: {
+                verificationLabel = 'Verified';
+                break;
+            }
+        }
+    }
```
cgero-eth, line 41: This can be simplified by using a ternary operator as we don't need the switch statement:
```
const contractClassName = action.inputData == null ? 'text-warning-800' : 'text-neutral-500':
const contractName = action.inputData?.contract;
const icon = action.inputData == null
  ? { className: 'text-warning-500', icon: IconType.WARNING }
  : { className: 'text-primary-300', icon: IconType.SUCCESS }
```
The `verificationLabel` should just be the `inputData.contract` which is the contract name the action is interacting with

src/modules/components/proposal/proposalActions/proposalActionsActionVerfication/proposalActionsActionVerification.stories.tsx
```diff
@@ -0,0 +1,36 @@
+import type { Meta, StoryObj } from '@storybook/react';
```
cgero-eth, line 1: Given that this is an internal component that is not exported, we don't need the stories for it

src/modules/components/proposal/proposalActions/proposalActionsTypes/proposalAction.ts
cgero-eth: Well, the `withdraw` action is a special case. In general, for all the actions, we're displaying the name of the smart contract the DAO is interacting with:
Add members => from: DAO to: Multisig
Mint token => from: DAO to: GovernanceToken
Not sure if we should have a special case for the `withdraw`, let's double check this with selim & evan

cgero-eth (approved):
LGTM, thanks @thekidnamedkd!

PR #262, opened 2024-08-05
fix(APP-3440): Fix NumberInput component to correctly update value on buttons click

cgero-eth:
## Description

- From version `7.6.0`, the `react-imask` does not correctly trigger the `onAccept` callback on value change, see issue here => https://github.com/uNmAnNeR/imaskjs/pull/1045

Task: [APP-3440](https://aragonassociation.atlassian.net/browse/APP-3440)

## Type of change

## Checklist:

        the latest version.

[APP-3440]: https://aragonassociation.atlassian.net/browse/APP-3440?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,10 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    ignore:
+      # From version 7.6.x, react-imask doesn't correctly trigger the `onAccept` callback when changing the value
+      # programmatically (see https://github.com/uNmAnNeR/imaskjs/pull/1045)
+      - dependency-name: "react-imask"
     groups:
       minor-and-patch:
         update-types:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 -   Fix ENS name truncation on `<Wallet />` module component
 -   Update `<Wallet />` module component to only resolve user ENS name when name property is not set
 -   Fix expand behaviour of `TextAreaRichText` core component when used inside a dialog and hide the input label
+-   Fix `NumberInput` component to correctly update values on plus / minus buttons click
 
 ### Added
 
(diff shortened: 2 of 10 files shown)


shan8851:
@cgero-eth all changes look good and everything is working as expected. 

However I wonder if we can slightly improve the UX. When we have `min` or `max` set we could disable the individual `+` or `-` button as right now it is clickable and nothing happens, so we should indicator somehow to the user that it is not possible by disabling buttons and/or having some type of label. My vote would just be to disable the buttons.

I don't think we should necessarily add this change to this PR though, seen as though right now input is not working properly the priority should be to get this fix merged, but perhaps one for the future. wdyt?

shan8851 (dismissed):

cgero-eth:
> @cgero-eth all changes look good and everything is working as expected.
> 
> However I wonder if we can slightly improve the UX. When we have `min` or `max` set we could disable the individual `+` or `-` button as right now it is clickable and nothing happens, so we should indicator somehow to the user that it is not possible by disabling buttons and/or having some type of label. My vote would just be to disable the buttons.
> 
> I don't think we should necessarily add this change to this PR though, seen as though right now input is not working properly the priority should be to get this fix merged, but perhaps one for the future. wdyt?

Good point, could you create a task in Jira so we can keep track of this suggestion?

shan8851:
> > @cgero-eth all changes look good and everything is working as expected.
> > However I wonder if we can slightly improve the UX. When we have `min` or `max` set we could disable the individual `+` or `-` button as right now it is clickable and nothing happens, so we should indicator somehow to the user that it is not possible by disabling buttons and/or having some type of label. My vote would just be to disable the buttons.
> > I don't think we should necessarily add this change to this PR though, seen as though right now input is not working properly the priority should be to get this fix merged, but perhaps one for the future. wdyt?
> 
> Good point, could you create a task in Jira so we can keep track of this suggestion?

Done - [APP-3499](https://aragonassociation.atlassian.net/browse/APP-3499)

PR #330, opened 2024-11-05
fix(APP-3626): Update ProposalVoting Container/Stage components to correctly update on stage and status changes

cgero-eth:
## Description

- The radix-ui accordion does not update the active item when changing the `defaultValue` property. In order to mark the next stage active as soon as the previous stage is advanced, we need to control the accordion value by using the `value` and `onValueChange` accordion properties. Same changes are needed for the `ProposalVotingStage` component to properly update the default voting-tab when a stage status changes from PENDING to ACTIVE

Task: [APP-3626](https://aragonassociation.atlassian.net/browse/APP-3626)

## Type of change

## Checklist:

        the latest version.

[APP-3626]: https://aragonassociation.atlassian.net/browse/APP-3626?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/modules/components/proposal/proposalVoting/proposalVotingContainer/proposalVotingContainer.tsx
+++ b/src/modules/components/proposal/proposalVoting/proposalVotingContainer/proposalVotingContainer.tsx
@@ -12,13 +12,17 @@ export interface IProposalVotingContainerProps extends ComponentProps<'div'> {
      */
     description: string;
     /**
-     * Active stage that will be expanded by default for multi-stage proposals.
+     * Active stage that will be expanded for multi-stage proposals.
      */
     activeStage?: string;
+    /**
+     * Callback called when the user selects a stage, to be used for expanding the current active stage for multi-stage proposals.
+     */
+    onStageClick?: (stage?: string) => void;
 }
 
 export const ProposalVotingContainer: React.FC<IProposalVotingContainerProps> = (props) => {
-    const { title, description, className, children, activeStage, ...otherProps } = props;
+    const { title, description, className, children, activeStage, onStageClick, ...otherProps } = props;
 
     const processedChildren = Children.toArray(children);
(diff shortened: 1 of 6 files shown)


src/modules/components/proposal/proposalVoting/proposalVotingContainer/proposalVotingContainer.tsx
```diff
@@ -30,7 +34,7 @@ export const ProposalVotingContainer: React.FC<IProposalVotingContainerProps> =
                 <p className="text-base font-normal leading-normal text-neutral-500">{description}</p>
             </div>
             {isMultiStage && (
-                <Accordion.Container isMulti={false} defaultValue={activeStage}>
+                <Accordion.Container isMulti={false} value={activeStage} onValueChange={onStageClick}>
                     {processedChildren.map((child, index) =>
                         React.isValidElement(child)
                             ? React.cloneElement(child, { ...child.props, index, isMultiStage })
```
thekidnamedkd, line 40: Maybe we set a key more explicitly here? Can we expect the stages to always come in the same order and not be subject to a reorder when using the string index + state as the control?

thekidnamedkd (commented):
Makes sense to me. Thanks for the update! Left one comment for musing.

src/modules/components/proposal/proposalVoting/proposalVotingContainer/proposalVotingContainer.tsx
cgero-eth: The proposal stages shouldn't get reordered but always follow the same order. As stage names can be duplicated, the `stageIndex` is the only unique value between stages at the moment

PR #371, opened 2024-12-16
feat(APP-3834): Remove 24h price change feature from AssetDataListItemStructure

shan8851:
## Description

Update AssetDataListItemStructure module component to remove priceChange property and percentage calculation

Task: [APP-3834](https://aragonassociation.atlassian.net/browse/APP-3834)

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] I have selected the correct base branch.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have made corresponding changes to the documentation.
- [x] My changes generate no new warnings.
- [ ] Any dependent changes have been merged and published in downstream modules.
- [x] I ran all tests with success and extended them if necessary.
- [x] I have updated the `CHANGELOG.md` file in the root folder of the package after the [UPCOMING] title and before the
      latest version.
- [ ] I have tested my code on the test network.

[APP-3834]: https://aragonassociation.atlassian.net/browse/APP-3834?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/modules/components/asset/assetDataListItem/assetDataListItemStructure/assetDataListItemStructure.tsx
```diff
@@ -68,43 +40,26 @@ export const AssetDataListItemStructure: React.FC<IAssetDataListItemStructurePro
...
-                        </>
-                    ) : (
-                        <span className="text-sm leading-tight text-neutral-800 md:text-base">
-                            {copy.assetDataListItemStructure.unknown}
-                        </span>
-                    )}
+        <DataList.Item
+            className={classNames('flex items-center justify-between gap-x-3 py-3 md:py-5', className)}
+            {...otherProps}
+        >
```
thekidnamedkd, line 47: Looks good and technically unrelated, but could we adjust the min-h on the skeleton down to 88.5 to match layout? The height on the tag was giving us a few extra pixels to compensate for.

src/modules/components/asset/assetDataListItem/assetDataListItemStructure/assetDataListItemStructure.test.tsx
```diff
@@ -20,24 +20,10 @@ describe('<AssetDataListItem.Structure /> component', () => {
...
         expect(screen.getByText(props.amount)).toBeInTheDocument();
-        expect(screen.getByText('+$2.50K')).toBeInTheDocument();
-        expect(screen.getByText('+25.00%')).toBeInTheDocument();
-    });
-
-    it('renders unknown with fiatPrice is not set', () => {
-        const props = { fiatPrice: undefined };
-        render(createTestComponent(props));
-        expect(screen.getByText('Unknown')).toBeInTheDocument();
-    });
```
thekidnamedkd, line 36: Is there a reason we are removing this fallback (and test) from the DLI as well?

thekidnamedkd (commented):
Thanks for the changes @shan8851! ✨

src/modules/components/asset/assetDataListItem/assetDataListItemStructure/assetDataListItemStructure.test.tsx
shan8851: As far as I can tell now that we don't use price change there is no need to use the fallback as we always have an amount and USD defaults to 0 if we don't have it.

thekidnamedkd: Fair point. Let's check maybe what product thinks. 0 feels like it's inaccurate for faulty data possibility, but 'Unknown' doesn't really match any other usage. Perhaps we split the difference and use the '-' fallback, but let's check.

shan8851: Actually no, the more I think about this you are correct, I have restored this with some small additions due to overall layout shift of the DLI

src/modules/components/asset/assetDataListItem/assetDataListItemStructure/assetDataListItemStructure.tsx
```diff
@@ -68,45 +43,44 @@ export const AssetDataListItemStructure: React.FC<IAssetDataListItemStructurePro
...
-                            <div className="flex items-center gap-x-1">
-                                <span className={changedAmountClasses}>{formattedPriceChanged}</span>
-                                <Tag label={formattedPriceChangedPercentage!} variant={tagVariant} />
-                            </div>
-                        </>
-                    ) : (
-                        <span className="text-sm leading-tight text-neutral-800 md:text-base">
-                            {copy.assetDataListItemStructure.unknown}
-                        </span>
+                    {!fiatPrice && (
```
cgero-eth, line 55: I believe the component changed, we're now displaying `unknown` as price if we don't have a token price, we should update the component to reflect this as well

thekidnamedkd (dismissed):
LGTM ✨

src/modules/components/asset/assetDataListItem/assetDataListItemStructure/assetDataListItemStructure.tsx
```diff
@@ -68,43 +43,26 @@ export const AssetDataListItemStructure: React.FC<IAssetDataListItemStructurePro
...
+            {...otherProps}
+        >
+            <div className="flex min-w-0 items-center gap-3">
+                <Avatar src={logoSrc} responsiveSize={{ md: 'md', sm: 'sm' }} className="block shrink-0" />
+                <span className="truncate text-base leading-tight text-neutral-800 md:text-lg">{name}</span>
+            </div>
+            <div className="flex min-w-0 items-end justify-end gap-x-2 text-right">
+                <div className="flex min-w-0 flex-col gap-y-1">
+                    <span className="shrink-0 text-base leading-tight text-neutral-800 md:text-lg">
+                        {fiatPrice ? formattedPrice : copy.assetDataListItemStructure.unknown}
```
cgero-eth, line 58: The formatter utils supports a fallback string, we could specify the `unknown` string directly there

src/modules/components/asset/assetDataListItem/assetDataListItemStructure/assetDataListItemStructure.tsx
```diff
@@ -68,43 +43,26 @@ export const AssetDataListItemStructure: React.FC<IAssetDataListItemStructurePro
...
-                    )}
+        <DataList.Item
+            className={classNames('flex items-center justify-between gap-x-3 py-3 md:py-5', className)}
+            {...otherProps}
+        >
+            <div className="flex min-w-0 items-center gap-3">
+                <Avatar src={logoSrc} responsiveSize={{ md: 'md', sm: 'sm' }} className="block shrink-0" />
+                <span className="truncate text-base leading-tight text-neutral-800 md:text-lg">{name}</span>
+            </div>
+            <div className="flex min-w-0 items-end justify-end gap-x-2 text-right">
```
cgero-eth, line 55: Are all `items-end`, `justify-end` and `text-right` classes required? We can probably just keep `text-right`

src/modules/components/asset/assetDataListItem/assetDataListItemStructure/assetDataListItemStructure.tsx
```diff
@@ -68,43 +43,26 @@ export const AssetDataListItemStructure: React.FC<IAssetDataListItemStructurePro
...
+                <div className="flex min-w-0 flex-col gap-y-1">
+                    <span className="shrink-0 text-base leading-tight text-neutral-800 md:text-lg">
+                        {fiatPrice ? formattedPrice : copy.assetDataListItemStructure.unknown}
+                    </span>
+                    <div className="flex min-w-0 items-center gap-1">
+                        <p className="shrink-0 text-sm leading-tight text-neutral-500 md:text-base">
+                            {formattedAmount}
+                        </p>
+                        <p className="min-w-0 truncate text-sm leading-tight text-neutral-500 md:text-base">{symbol}</p>
```
cgero-eth, line 64: Looks like without `min-w-0` the text is still truncated, could you double check the classes added to div & p tags to avoid adding classes that aren't needed?

src/modules/components/asset/assetDataListItem/assetDataListItemStructure/assetDataListItemStructure.tsx
shan8851: @cgero-eth I actually tried this first, but it didn't work, I guess because we default fiat amount  to 0, but we can safely remove that actually as it will just return NaN in the fiatAmount var and use the fallback

PR #717, opened 2026-07-04
fix: Set explicit text color on proposal status context label

evanaronson:
## Description

The status context label on `ProposalDataListItem.Structure` (the **X** in "in X" next to the status tag) rendered with no text color class, so it inherited the consuming app's default color — plain black instead of the design system's `text-neutral-800`.

This adds `text-neutral-800` to the wrapping element.

An audit of all 150 component source files in the kit found no other text nodes missing an explicit color class — this was the only offender.

## Checklist
- [x] Changeset added
- [x] Component tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)


thekidnamedkd (approved):
lgtm 👍

PR #719, opened 2026-07-12
fix(APP-989): Allow empty strings in action ABI string inputs

harryburger:
## Description

Fixes [APP-989](https://linear.app/aragon/issue/APP-989/allow-empty-strings-in-action-abi-input): the action composer rejected empty strings in ABI inputs, blocking actions where an empty Solidity string is the intended value (e.g. `functionSignatures: ["", "", ""]`, whose own ABI notice describes empty as valid).

**Root cause:** every decoder field was validated with a hardcoded `required: true`, and the required check (`value.toString().length > 0`) can never pass for an empty string — so `string` parameters, including `string[]` elements (which render as individual `string` leaf fields), could never be submitted empty.

**Changes:**

- `ProposalActionsDecoderTextFieldEdit`: extend the existing array validation-skip to `string` and `tuple` types. `string` fields have nothing to validate (any value, including empty, is a valid Solidity string — and with `required` off, the validator could only ever return "valid"). `tuple` fields are hidden registration fields whose components validate themselves visibly; keeping validation on them meant a `tuple(string)` holding an empty string failed on an invisible field, silently blocking submit with no error shown. Leaf types (`address`, `bool`, `bytes*`, `uint*`) still validate with `required: true` — notably `uint`, whose `^[0-9]*$` regex would accept an empty string if the required check were relaxed instead of skipped.
- `ProposalActionsDecoderTextFieldEdit`: initialise `string` fields to `''` (mirroring the existing `[]` init for arrays) so untouched empty elements are present in the form value and `encodeFunctionData` produces calldata matching what is on screen, instead of dropping them.
- `ProposalActionsDecoderField`: fix the array-row layout shift from the ticket. The validation alert renders below the input in the same flex row, so with `items-center` the `[i]:` label and remove button re-centered (jumped down) whenever an alert appeared. Rows now use `items-start` and center the label and button against the input's fixed 48px height with constant offsets (`flex h-12 items-center` on the label, `mt-2` on the 32px button) — same look, but nothing moves when an alert shows. Rows without a 48px text input (nested tuple/array boxes, boolean radio groups on edit mode) skip the offsets and keep plain top alignment.
- New `StringArrayType` story covering a `string[]` parameter with empty elements.

**How to test:**

- Storybook: `ProposalActions.Decoder` → `StringArrayType` story — clear the string elements, no "required" error should appear and the `X` buttons should not shift.
- In the app, against a Sepolia DAO: add a custom action for ENS `StaticBulkRenewal` at `0x6394b694a8C0DC716e447802E568F0Fb2c4E0965` (verified on Sepolia Etherscan) and select `renewAll(string[] names, uint256 duration, bytes32 referrer)`. Empty elements in `names` should be accepted, while leaving `duration` or `referrer` blank should still show a required error — covering both the relaxed and the still-validated paths in one function.

## Type of change

## Developer Checklist:

## Review Checklist:

### DIFF
--- a/src/modules/components/proposal/proposalActions/proposalActionsDecoder/proposalActionsDecoderTextField/proposalActionsDecoderTextFieldEdit.test.tsx
+++ b/src/modules/components/proposal/proposalActions/proposalActionsDecoder/proposalActionsDecoderTextField/proposalActionsDecoderTextFieldEdit.test.tsx
@@ -1,6 +1,7 @@
 import { render, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import * as ReactHookForm from 'react-hook-form';
+import { modulesCopy } from '../../../../../assets';
 import * as useFormContext from '../../../../../hooks';
 import {
     type IProposalActionsDecoderTextFieldEditProps,
@@ -167,7 +168,9 @@ describe('<ProposalActionsDecoderTextFieldEdit /> component', () => {
         const parameter = { name: 'uintParam', type: 'uint256', value: undefined };
         render(createTestComponent({ parameter }));
         const { validate } = useControllerSpy.mock.calls[0][0]!.rules!;
-        expect((validate as (value: unknown) => unknown)('')).toEqual(expect.any(String));
+        expect((validate as (value: unknown) => unknown)('')).toEqual(
+            modulesCopy.proposalActionsDecoder.validation.required(parameter.name),
+        );
     });
 
     it('initialises string types to empty strings when value is null', () => {
--- a/src/modules/components/proposal/proposalActions/proposalActionsDecoder/proposalActionsDecoderTextField/proposalActionsDecoderTextFieldEdit.test.tsx
+++ b/src/modules/components/proposal/proposalActions/proposalActionsDecoder/proposalActionsDecoderTextField/proposalActionsDecoderTextFieldEdit.test.tsx
@@ -90,8 +90,8 @@ describe('<ProposalActionsDecoderTextFieldEdit /> component', () => {
         });
     });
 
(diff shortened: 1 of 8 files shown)


arabot-1:
🚀 **Preview Deployment:** [View Here](https://gov-ui-duryllcp9-aragon-app.vercel.app)

src/modules/components/proposal/proposalActions/proposalActionsDecoder/proposalActionsDecoderTextField/proposalActionsDecoderTextFieldEdit.test.tsx
```diff
@@ -163,6 +163,43 @@ describe('<ProposalActionsDecoderTextFieldEdit /> component', () => {
         );
     });
 
+    it('rejects empty values for uint types', () => {
+        const parameter = { name: 'uintParam', type: 'uint256', value: undefined };
+        render(createTestComponent({ parameter }));
+        const { validate } = useControllerSpy.mock.calls[0][0]!.rules!;
+        expect((validate as (value: unknown) => unknown)('')).toEqual(expect.any(String));
```
thekidnamedkd: Tiny thought: we could assert the actual required error message here instead of `expect.any(String)` — makes it a little more obvious it's the required check catching the empty value and not the number regex.

thekidnamedkd (dismissed):
LGTM! Took it for a spin on the preview deployment too: empty `string[]` elements go through cleanly and encode as `""` in the calldata, `address`/`uint` fields still show their required errors, and the array rows stay put when an alert pops in. Nice one 🎉

thekidnamedkd (approved):
lgtm 🔥

PR #745, opened 2026-08-12
fix(APP-1011): Repair the published CSS bundle and literal number affixes

thekidnamedkd:
## Description

Two consumer-facing defects found by the [app design-sync run](https://github.com/aragon/app/pull/1243), plus the missing CI gate that let the first one ship twice. The `min`/`max` clamping decision is deliberately **not** here — it's in the stacked PR, because it's a behaviour change under discussion and these two are not.

### 1. `build.css` ships corrupted selector lists

`rollup-plugin-postcss`'s `minimize: true` runs cssnano 5, which predates CSS nesting. Tailwind v4 emits every variant utility as a nested block, and the merge pass collapsed unrelated selector lists onto one nested body. Reproduced on `main` at `6f8ac008` (v2.9.0) — the report was filed against 2.8.1, but it never got fixed:

| selectors merged | inherited child rule |
|---:|---|
| 140 `2xl:*` | `& > :last-child { border-style: none }` |
| 128 `md:*`/`sm:*` | `:where(& > :not(:last-child)) { margin-block: 0 }` |
| 54 `md:*` | `:is(& > *) { border-right-width: 1px }` |
| 7 `md:*` | `:where(& > :not(:last-child)) { margin-inline… }` |

Row 3 is the stray vertical bar on `AlertCard`/`Accordion`. Invisible to us because the app and Storybook compile from source; broken for every consumer of the published file.

**Root cause is the missing check, not the merge:** nothing in this repo has ever rendered or parsed `build.css`, so a corrupted bundle had no way to fail. Fix is to minify with Tailwind's own optimizer (Lightning CSS, nesting-aware) and assert in CI that the bundle equals a source compile — `pnpm check:css`, 33 lines, no heuristics. Verified it fails on a cssnano build (`103017 vs 118809 chars`) and passes on this one.

The bundle grows 103 KB → 119 KB. That difference is the rules that were previously being merged away.

### 2. `prefix`/`suffix` are parsed as an imask pattern

`useNumberMask` interpolated them into the mask, where `0`, `a` and `*` are definitions, so any affix containing one lost that character. `suffix="aUSDC"` rendered `_USDC`. Live in the app — `tokenMintTokensAction` and `proposalCreationEligibilityField` both pass `suffix={tokenSymbol}`, so Aave-style aTokens are affected.

Tests for this run the real imask instance; the existing suite mocks `useNumberMask`, which is why a pattern bug could never surface. Confirmed the new cases fail on the pre-fix hook.

## Type of change

## Developer Checklist:

- [x] Manually smoke tested the functionality locally — `pnpm build` then `pnpm check:css`, both directions
- [x] Confirmed there are no new warnings or errors in the browser console — no UI change
- [x] Made the corresponding changes to the documentation — changeset
- [x] Confirmed there are no new warnings on automated tests — 1122 pass
- [x] Selected the correct base branch (`main`)

## Notes

- Touches `package.json` scripts and `library-test.yml`, same two places as #739 (APP-723). Whichever lands second rebases; one-line conflict.
- Replaces #742, which mixed these fixes with the `min`/`max` change.
- The durable version of check 1 is a rendered baseline diff — APP-733 (Playwright + Argos) is its home. `check:css` only proves the file is internally consistent, not that a browser paints it correctly.


milosh86 (approved):
looks good 👍

PR #746, opened 2026-08-12
fix(APP-1011): Clamp out-of-range number input values

thekidnamedkd:
## Description

**Draft: the behaviour decision is open and tracked in APP-1083.** #745 has merged, so this now targets `main` directly and the diff is the clamp alone.

`InputNumber` turns an out-of-range number into a *different* number. imask rejects the character that breaches the bound and keeps the longest in-range prefix:

| input | `max` | renders | `onChange` emits |
|---|---|---|---|
| types `60` (duration minutes) | 59 | `6` | `"6"` |
| types `25` (duration hours) | 23 | `2` | `"2"` |
| types `101` (percentage) | 100 | `10` | `"10"` |
| receives `value="250"` | 100 | `25` | `"25"` |

Typed, pasted and controlled values alike. Affected callsites in the app: `advancedDateInputDuration` (59, 23), `setupStrategyDialogDistributionRecipientItem` (100), `gaugeVoterVoteDialogItem` (`WEIGHT_MAX`), `tokenProposalCreationSettings` (computed), and `numberProgressInput` (`max={total}`, dynamic).

Everyone agrees the current output is wrong. The open question is which correct behaviour replaces it:

- **A (this PR)** — clamp to the boundary. One line, no API change, and every caller already treats `max` as a bound. Cost: an out-of-range error state becomes impossible with `max` set; it has to come from the `alert` property.
- **B** — drop the bounds from the mask and validate in forms. Never alters input, error states work naturally. Cost: validation must be added at five callsites, two of which feed transaction parameters (gauge vote weight, distribution percentages), so shipping it without that work is a safety regression.

A doesn't foreclose B, and an opt-out prop is deliberately not added — nothing needs it yet.

### One thing to settle

`numberProgressInput` passes a dynamic `max={total}`. Under A, a shrinking `total` rewrites a value the user already entered and fires `onChange` from a data update rather than a user action. Either accept that, or have that one component drop `max` and validate instead — app-side, one file.

### `min` is not symmetric, whichever option wins

Bounds cannot block partial input: `5` must be typeable on the way to `50` in a `min={10}` field. Measured — `min={10}` with `value="5"` renders `10` (committed values are raised), but typing is unrestricted, and `min={0}` with `value="-5"` renders `5`, dropping the sign rather than clamping to `0`. So "min/max are hard bounds" is only true of `max`. The sign case is pre-existing, same root cause, and left alone: fixing it means special-casing the sign for a contrived input. Docs now state what actually happens.

## Type of change

Behaviour change worth calling out: over-max input clamps rather than losing a digit.

## Developer Checklist:

- [x] Manually smoke tested the functionality locally — jsdom probes for typed, pasted and controlled values, and for `min` in both signs
- [x] Confirmed there are no new warnings or errors in the browser console — no UI change
- [x] Made the corresponding changes to the documentation — `min`/`max` props on both components, `MinMax` story, changeset
- [x] Added tests that prove my fix is effective or that my feature works — confirmed failing pre-fix
- [x] Confirmed there are no new warnings on automated tests — 1124 pass
- [x] Selected the correct base branch — stacked on #745 by design


milosh86 (approved):
Seems reasonable and it feels good when testing 👍

PR #762, opened 2026-08-26
fix(APP-1089): Stop AddressOutput nesting controls in interactive containers

thekidnamedkd:
## Description

`AddressOutput` defaults `copy` and `reveal` to `true`, so a callsite that sits inside an interactive container has to switch them off. Six callsites shipped in 2.10.0 without doing that, or doing half of it.

Live DOM from the `AssetTransfer` → `Default` story before this change:

```
<a href="https://etherscan.io/address/0xd8dA…6045">
  ├── <button>vitalik.eth</button>   ← reveal trigger
  └── <button aria-label="Copy">     ← Clipboard
```

Keyboard order per row was `A → BUTTON → BUTTON`: two phantom tab stops inside one row link, and Enter on the reveal button does nothing. The reveal button has no click handling, only `onPointerDown`, so a click bubbles and the row navigates while a touch tap pins the tooltip *and* navigates.

`memberDataListItemStructure`, `daoDataListItemStructure` and `voteDataListItemStructure` each carried the comment "the reveal must not add a nested interactive trigger" directly above `reveal={true}`. This passes `reveal={!isInteractiveItem}` to match the `copy` line above it. `transactionDataListItemStructure` and `assetTransferAddress` always render inside a link, so both affordances are off there, and the `Wallet` handle sits inside the connect `<button>`.

`proposalDataListItemStructure` and `smartContractFunctionDataListItemStructure` are untouched: their `<object type="unknown">` wrapper is the deliberate [kizu.dev](https://kizu.dev/nested-links/) isolation, so their copy and reveal are intentional.

Verified in Storybook, live DOM:

| Surface | Before | After |
| -- | -- | -- |
| `AssetTransfer` → `Default` | 2 anchors × `[BUTTON reveal, BUTTON Copy]` | 2 anchors, no nested controls |
| Transaction `ExecutionByAddress` | `[BUTTON 0x1234…1234, BUTTON Copy]` | no controls, address is plain text |
| `Wallet` → `Address` | copy circle + reveal button inside the button | 1 button, no tooltip on hover |

Follow-up: #APP-1108 asks whether `DataList.Item` should stop swallowing the row at all, which would let these controls come back legitimately. A second PR resolves the defaults from the container so a future callsite cannot get this wrong.

Linear: https://linear.app/aragon/issue/APP-1089

## Type of change

## Developer Checklist:

- [ ] Made the corresponding changes to the documentation
- [ ] Added tests that prove my fix is effective or that my feature works
- [x] Reviewed that the files changed in GitHub’s UI reflect my intended changes

No tests here on purpose: every line is a prop value whose behaviour belongs to `AddressOutput`, already covered in `addressOutput.test.tsx`. A `Wallet` test asserting "no Copy button" would be testing `AddressOutput`'s `copy` prop through a second component. The follow-up PR adds tests where the real branching lives.

Removing the nested `<button>` from the `Wallet` button also clears the `<button> cannot contain a nested <button>` DOM nesting warning that `wallet.test.tsx` was printing.


milosh86 (approved):
👍

PR #763, opened 2026-08-26
feat(APP-1089): Resolve AddressOutput controls from the container

thekidnamedkd:
## Description

Whether a nested interactive control is legal is a fact about the *container*, not the caller. Today every callsite re-derives it — `('href' in otherProps && otherProps.href != null) || otherProps.onClick != null`, copy-pasted into three files, and all three then passed `reveal={true}` anyway. This moves the fact to the two chokepoints that already exist and makes the props an explicit override.

```tsx
// core/components/interactiveAncestor
const interactiveAncestorContext = createContext(false);
export const InteractiveAncestorProvider = interactiveAncestorContext.Provider;
export const useInteractiveAncestor = () => useContext(interactiveAncestorContext);

// addressOutput.tsx
const showCopy = copy ?? !hasInteractiveAncestor;
const showReveal = reveal ?? !hasInteractiveAncestor;
```

`LinkBase` provides it, which covers every anchor in the kit including the `DataList.Item` link branch, `AssetTransferAddress` and `smartContractFunctionDataListItemStructure`. `DataListItem`'s interactive-div branch provides it for `role="button"` rows. `Wallet` provides it around its connect button, which a container cannot restructure its way out of.

Unset means "do the right thing here", so `ProposalDataListItem` and `SmartContractFunctionDataListItem` pass `copy` and `reveal` explicitly to keep their `<object>`-isolated links working — React context flows straight through `<object>`, so the escape hatch has to be stated in props. That override is also what consumers outside the kit get: `InteractiveAncestorProvider` is exported so app-side row containers can mark themselves.

Net effect on the kit is deletion: `isInteractiveItem` and its comment leave member, DAO and vote; the explicit `copy`/`reveal` props leave four more callsites.

Verified in Storybook, live DOM, all four container shapes:

| Story | Controls rendered | Nested in anchor/button |
| -- | -- | -- |
| `AssetTransfer` → `Default` | 2 row anchors only | none |
| `Wallet` → `Address` | 1 button | none |
| `AddressOutput` → `Default` (standalone) | reveal + Copy | none |
| `SmartContractFunctionDataListItem` → `Verified` | link + link + Copy | Copy, unchanged and deliberate |

Semver is minor, not major: standalone consumers resolve to `true` exactly as today, and only the nested-in-interactive case changes, which is the case #APP-1108 and the sibling fix PR call a bug. A consumer who deliberately put copy inside a link row keeps it by passing the prop.

Two things worth a reviewer's attention:

1. I considered a dev-only warning for components that are not converted, and dropped it. It would fire on `Clipboard` inside `Link`, which the kit supports on purpose (`clipboard.stories.tsx` → `InsideLink`). Telling accidental nesting from deliberate needs a per-component policy, and a component with a policy can simply not render the control.
2. The `smartContract` Copy button is still a `<button>` inside an `<a>`. This PR preserves 2.10.0 behaviour exactly; the `<object>` trick is documented for nested *links* and I doubt it launders a nested button. That belongs to #APP-1108 rather than a silent change here.

Related: https://linear.app/aragon/issue/APP-1108 asks whether the row should stop swallowing its contents at all, which would let these controls return legitimately. Depends on the sibling fix PR only for merge order — both touch the same prop lines, so whichever lands second drops the redundant deletions.

Linear: https://linear.app/aragon/issue/APP-1089

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## Developer Checklist:

- [x] Reviewed that the files changed in GitHub’s UI reflect my intended changes

Three tests in `addressOutput.test.tsx` cover the resolution branches through real `DataList.Item` containers — link, clickable row, and explicit override — so they fail if either provider or the `??` resolution breaks. `pnpm build` and `pnpm css:check` pass with the new public export, confirmed in `dist/types/src/core/components/interactiveAncestor/`.

### DIFF
--- a/.changeset/interactive-ancestor-context.md
+++ b/.changeset/interactive-ancestor-context.md
@@ -1,0 +1,5 @@
+---
+"@aragon/gov-ui-kit": minor
+---
+
+`AddressOutput` now takes its `copy` and `reveal` defaults from its container: both stay `true` standalone and become `false` inside an interactive element such as a link, an interactive `DataList.Item` row, or the `Wallet` button, where a nested control is invalid markup and competes with the container for the click. Pass either prop explicitly to override, as `ProposalDataListItem` and `SmartContractFunctionDataListItem` do for their isolated links. Containers built outside the kit can mark themselves with the newly exported `InteractiveAncestorProvider`.
--- a/src/core/components/dataList/dataListItem/dataListItem.tsx
+++ b/src/core/components/dataList/dataListItem/dataListItem.tsx
@@ -1,5 +1,6 @@
 import classNames from 'classnames';
 import type { AnchorHTMLAttributes, HTMLAttributes } from 'react';
+import { InteractiveAncestorProvider } from '../../interactiveAncestor';
 import { LinkBase } from '../../link';
 
 export type DataListItemVariant = 'primary' | 'select';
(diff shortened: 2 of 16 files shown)


milosh86 (commented):
I'm not sure about this approach tbh, I need to think about it more.

PR #765, opened 2026-08-27
fix(APP-1089): Restore address reveal in interactive rows

thekidnamedkd:
## Description

Preserve the full `AddressOutput` interaction model inside app-owned interactive surfaces.

Interactive data-list rows, asset-transfer address cards, and the wallet control render their row action as a sibling overlay instead of an ancestor of the address output. Whole-row links/actions remain intact while address reveal and copy controls stay available without nested interactive elements.

`DefinitionList.Item` delegates on-chain entity link, reveal, and copy behavior to its child through `link.isOnchainEntity`.

Long tooltip values are constrained to Radix's available viewport width, wrap at character boundaries, and render above avatar/content layers on mobile.

## Type of change

- [x] Bug fix
- [x] New feature: additive `DefinitionList.Item` link metadata

## Verification

- [x] Full Jest suite: 167 suites, 1222 tests passed
- [x] Type-check passed
- [x] Lint check passed
- [x] Storybook build passed
- [x] Package build passed
- [x] Browser smoke check confirmed interactive member rows expose reveal and Copy controls without nesting them in the row link
- [x] Mobile browser check confirmed long address tooltips wrap within the viewport and stack above avatars

## Notes

The existing `hasInteractiveAncestor` escape hatch remains available for consumers that intentionally need passive behavior; app callsites no longer suppress AddressOutput controls.


milosh86 (approved):
Looks good 👍

PR #767, opened 2026-08-27
fix: keep address copy controls inside cards

thekidnamedkd:
## Description

Keep long ENS/address labels and their copy controls inside member cards.

The shared `Clipboard` wrapper and `DataList.Item` interactive content now expose shrink boundaries, while `AddressOutput` renders its truncatable label as a constrained inline block. This lets the label ellipsize before the copy control instead of pushing the control outside the card.

## Type of change

- [x] Bug fix

## Verification

- [x] Full Jest suite: 167 suites, 1223 tests passed
- [x] Type-check passed
- [x] Lint check passed
- [x] Storybook build passed
- [x] Package build passed
- [x] Browser smoke check confirmed `michiganblockchain.eth` truncates and keeps Copy inside the card in the fixed-width three-column story.

### DIFF
--- a/.changeset/keep-copy-in-card.md
+++ b/.changeset/keep-copy-in-card.md
@@ -1,0 +1,5 @@
+---
+"@aragon/gov-ui-kit": patch
+---
+
+Keep address copy controls inside narrow cards.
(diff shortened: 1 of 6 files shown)


milosh86 (approved):
👍

PR #768, opened 2026-08-27
refactor: standardize on-chain entity address rendering

thekidnamedkd:
## Description

`AddressInput` and `AddressOutput` were independent sibling components under `src/modules/components/address`; neither rendered nor depended on the other. The shared barrel grouped two different concerns: `AddressInput` handles forms, wagmi, and ENS resolution, while `AddressOutput` displays an on-chain identifier.

This PR makes `AddressOutput` a core display primitive and makes `DefinitionList.Item` responsible for standard on-chain value rendering. It also fixes the interactive-container behavior exposed while migrating the callsites.

```tsx
<DefinitionList.Item term="Gauge address" link={{ href: gaugeAddressLink, isOnchainEntity: true }}>
    {gaugeAddress}
</DefinitionList.Item>
```

## What changed

- Moved `AddressOutput` and `addressUtils` to `src/core`; `AddressInput` remains in `src/modules`.
- `DefinitionList.Item` renders string children through `AddressOutput` when `link.isOnchainEntity` is set, so the item owns link, reveal, and copy behavior without duplicated child wrappers or link props.
- `copyValue` can supply the underlying on-chain value while the string child remains a custom label.
- Default `AddressOutput` labels truncate addresses and 32-byte hashes.
- `AddressOutput` and `Clipboard` use inline-safe roots.
- Added `InteractiveAncestorContext` so buttons, links, and clickable rows mark one boundary for their subtree. Nested outputs keep passive hover reveal without adding copy buttons, tab stops, or competing click targets.
- Updated interactive `DataList.Item` rows to preserve consumer flex/grid layout, keep child controls above the row overlay, and contain their stacking within the row.
- Updated `Wallet` to remain one interactive control with a passive, vertically aligned address label.
- Migrated module callsites to the core barrel. Root package exports remain unchanged.

## Type of change

- [x] Refactor (non-breaking; public exports unchanged)
- [x] Bug fix

## Verification

- [x] Full Jest suite: 167 suites / 1227 tests
- [x] Type-check, lint, and package build
- [x] Browser smoke in Storybook: on-chain item, transaction summary, moved `AddressOutput` stories, and connected `Wallet` interaction/alignment
- [x] Downstream app type-check against the packed tarball


milosh86 (approved):
All looks good now 🙌

PR #775, opened 2026-09-04
fix(wallet): keep address label hidden below md breakpoint

thekidnamedkd:
## Problem

After #768, the `Wallet` pill rendered its address label at every breakpoint. The small layout is supposed to be avatar-only.

`Wallet` passed `className="hidden md:block truncate"` to `AddressOutput`, and #768 started forwarding that `className` to the component root. `AddressOutput`'s own base `inline-flex` and the incoming `hidden` are both single-class utilities, so `classNames` order does not decide the winner — CSS source order does, and `inline-flex` is emitted after `hidden` in Tailwind's layer. The label stayed visible below `md`, the loading skeleton behaved the same way, and the pill kept its `md` padding at small sizes.

## Change

- `AddressOutput` label: `hidden md:block truncate` → `truncate max-md:hidden`. A `max-*` variant carries a media query, so it wins below `md` regardless of source order and stops fighting the base `inline-flex`.
- `StateSkeletonBar`: `block max-md:hidden`, matching the label's responsive visibility while retaining block dimensions.
- Pill padding: `p-1 md:pl-4` so the icon-only small layout keeps its square padding.

## Verification

- Storybook `Wallet` at 767/768px: label hidden with a 50px pill below `md`; label visible with a ~161px pill at `md` and above.
- `pnpm test -- --runInBand src/modules/components/wallet/wallet.test.tsx` (9 passed), `pnpm lint:check`, `pnpm type-check`, `pnpm build`, `pnpm css:check`.

### DIFF
--- a/src/modules/components/wallet/wallet.tsx
+++ b/src/modules/components/wallet/wallet.tsx
@@ -50,7 +50,7 @@ export const Wallet: React.FC<IWalletProps> = (props) => {
                     id={contentId}
                 >
                     {!user && copy.wallet.connect}
-                    {user && isEnsLoading && <StateSkeletonBar className="hidden md:block" size="lg" width={56} />}
+                    {user && isEnsLoading && <StateSkeletonBar className="block max-md:hidden" size="lg" width={56} />}
                     {user && !isEnsLoading && (
                         <AddressOutput
                             address={user.address}
--- a/src/modules/components/wallet/wallet.tsx
+++ b/src/modules/components/wallet/wallet.tsx
@@ -33,7 +33,7 @@ export const Wallet: React.FC<IWalletProps> = (props) => {
         'flex max-w-48 cursor-pointer items-center gap-3 rounded-full border border-neutral-100 bg-neutral-0 text-neutral-500 transition-all',
         'focus-ring-primary hover:border-neutral-200 active:bg-neutral-50 active:text-neutral-800 disabled:cursor-default',
         { 'px-4 py-2.5': user == null },
-        { 'p-1 xl:pl-4': user != null },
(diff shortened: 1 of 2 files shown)


src/modules/components/wallet/wallet.tsx
```diff
@@ -50,11 +50,11 @@ export const Wallet: React.FC<IWalletProps> = (props) => {
                     id={contentId}
                 >
                     {!user && copy.wallet.connect}
-                    {user && isEnsLoading && <StateSkeletonBar className="hidden xl:block" size="lg" width={56} />}
+                    {user && isEnsLoading && <StateSkeletonBar className="hidden md:block" size="lg" width={56} />}
                     {user && !isEnsLoading && (
                         <AddressOutput
                             address={user.address}
-                            className="hidden truncate xl:inline-flex"
+                            className="truncate max-md:hidden"
```
milosh86, line 57: it seems that we use 2 different approaches for the same thing here:

`hidden md:block` vs `max-md:hidden`

let's choose one

milosh86: to clarify `hidden md:block` is above in the `<StateSkeletonBar`

thekidnamedkd: Good catch. Updated `StateSkeletonBar` to `block max-md:hidden`, so both the skeleton and address label use `max-md:hidden`. The explicit `block` keeps the skeleton width and height effective at `md` and above. Re-ran the Wallet tests, lint, type-check, build, CSS check, and the 767/768px Storybook breakpoint check.

milosh86 (approved):

PR #778, opened 2026-09-08
fix(APP-1150): Stop tooltips from opening on focus that no key press preceded

milosh86:
## Description

Opening a dialog that contains an `AddressOutput` also opened the address tooltip, with nothing hovered.

**Root cause** — two Radix behaviours colliding:

1. A dialog moves focus into its content on open. Radix's `FocusScope` focuses the first tabbable element it finds, skipping links (`removeLinks`).
2. `@radix-ui/react-tooltip` opens the tooltip on *any* focus, with no way to tell a keyboard `Tab` from focus moved by script:
   ```js
   onFocus: composeEventHandlers(props.onFocus, () => {
     if (!isPointerDownRef.current) context.onOpen();
   })
   ```

So whenever the dialog's first tabbable element is one of the tooltip triggers `AddressOutput` renders, that tooltip is already open when the dialog appears. This is why it looked intermittent: a dialog whose header has a close button is fine, because focus lands there instead.

It affects `Clipboard` too — with `href` set, the reveal trigger is a link and gets skipped, so the `Copy` tooltip opens instead.

## Fix

The gate lives in `Tooltip`, so every tooltip in the kit is covered.

**1. `Tooltip` refuses an unwanted open rather than default-preventing the focus event.** It now always controls Radix's `open`, and honours an open request only when the pointer is over the trigger (a hover) or the keyboard just moved focus:

```ts
if (nextOpen && !isPointerOverTriggerRef.current && !getIsKeyboardModality()) {
    return;
}
```

Refusing at the open, rather than on the event, is deliberate. React shares one event object along the whole propagation path and every Radix primitive composes its handlers with `composeEventHandlers`, which skips them once the event is default-prevented — so default-preventing the focus would silently disable the focus handling of any trigger passed through `triggerAsChild`. A `Dropdown.Item` wrapped in a `Tooltip` would stop highlighting, and the same mechanism reaches `RovingFocusGroupItem`, `Tabs.Trigger` and `RadioGroupItem`. There is now a test for that.

Pointer-over is tracked with `onPointerMove`/`onPointerLeave` on the trigger, ignoring touch pointers to match Radix's own hover logic. `defaultOpen` seeds the internal state instead of being passed to Radix.

**2. `keyboardModality.ts`** (new, internal — not exported from the public barrel, so the API surface is unchanged) tracks whether the keyboard just *moved focus*, via capture-phase `keydown`/`pointerdown` listeners on `document`.

Only keys whose default action moves focus count — `Tab`, the arrows, `Home`/`End`, `PageUp`/`PageDown` — and an activation key clears the flag:

```ts
isKeyboardModality = focusMovingKeys.has(event.key);
```

`Enter` and `Space` press a control; they do not move focus. Any focus that follows one of them was placed by script, which is exactly what a dialog does to its first tabbable element. Without the clearing, an earlier `Tab` would still be standing when the dialog opened.

The listeners are installed once for the page lifetime and deliberately never torn down: they have to observe the interaction that *precedes* a focus, which happens while no tooltip is mounted yet whenever a dialog owns the only tooltips on the page. Tying them to a mounted tooltip leaves the modality stale for exactly that case. A `typeof document !== 'undefined'` guard keeps importing the kit on a server free of side effects.

**3. `AddressOutput` loses `suppressOpenRef`.** Its only job was suppressing the open that a tap-induced focus caused, which the gate now refuses at the source. Keeping it would have been actively harmful: it relied on a subsequent open arriving to clear itself, so it went stale and swallowed the next legitimate reveal after a tap on a link address.

### A note on `:focus-visible`

This intentionally does **not** match the CSS pseudo-class. A dialog opened with `Enter` leaves its autofocused trigger matching `:focus-visible` natively, but showing a tooltip nobody asked for is wrong regardless of input method, so the gate refuses it. Real keyboard navigation into a trigger still reveals normally.

The pseudo-class could not have been used directly anyway: browsers only expose it as a style, and **jsdom aliases it to `:focus`** — `matches(':focus-visible')` returns `true` even for programmatic focus, which would have made the distinction untestable.

## Verification

29 new tests. 1256 pass in total.

**8 confirmed to fail without the gate** (removed and re-run), so they are genuine regression coverage:

- `addressOutput.test.tsx` — a dialog autofocusing the reveal button, the copy control (address as a link), and the copy control alone.
- `addressOutput.test.tsx` — a dialog opened with `Enter` and with `Space`.
- `addressOutput.test.tsx` — a dialog that owns the only tooltips on the page, covering the mount/unmount cycle.
- `clipboard.test.tsx` — a dialog autofocusing the copy control directly, rather than transitively through `AddressOutput`.
- `tooltip.test.tsx` — a primitive-level focus with no preceding key press.

The rest guard what must keep working: a keyboard `Tab` still opens the tooltip, a `Dropdown.Item` passed through `triggerAsChild` keeps its `data-highlighted` state, a tap on a link address still reveals on a later hover, and 18 unit tests for `keyboardModality.ts` covering which keys set, clear and leave the modality untouched.

### Browser checks

jsdom cannot answer everything here, so the following were verified in a real browser against the new `InsideDialog` story:

| Interaction | Tooltip |
| --- | --- |
| Dialog opened by mouse click | closed ✅ |
| Dialog opened with `Enter` | closed ✅ |
| Dialog opened with `Space` | closed ✅ |
| `Tab` / `Shift+Tab` onto a trigger | opens ✅ |
| Mouse hover | opens ✅ |

A **real touch tap** on a link address was traced under touch emulation, since removing `suppressOpenRef` depends on it:

```
pointerdown:touch → pointerup:touch → mousemove → mousedown → focusin → click:touch
```

Mobile browsers do fire an emulated mouse event after a tap, but it is a `mousemove`, not a `pointermove`, so `onPointerMove` never fires and the trigger is never treated as hovered. The anchor takes focus and the reveal stays closed.

`pnpm lint:check`, `pnpm type-check`, `pnpm test`, `pnpm build` and `pnpm css:check` all pass.

The new `InsideDialog` story doubles as the reproduction: pick which element ends up first in the dialog's tab order and open it with the mouse.

## Type of change

## Developer Checklist:

- [ ] (For User Stories only) Double-checked that all Acceptance Criteria are satisified
- [ ] Reviewed that the files changed in GitHub’s UI reflect my intended changes

## Review Checklist:


src/core/components/tooltip/useKeyboardModality.ts
```diff
@@ -0,0 +1,55 @@
...
+        // The listeners are shared by every mounted tooltip and capture the interaction before it reaches the trigger.
+        if (subscriberCount === 1) {
+            document.addEventListener('keydown', handleKeyDown, true);
+            document.addEventListener('pointerdown', handlePointerDown, true);
+        }
+
+        return () => {
+            subscriberCount -= 1;
+
+            if (subscriberCount === 0) {
```
thekidnamedkd: The flag outlives the listeners here, so the bug comes back:

1. tab to a tooltip → `true`
2. all tooltips unmount → listeners gone
3. click a button that opens a dialog → `pointerdown` unobserved
4. dialog mounts a trigger + autofocuses → still `true` → tooltip opens

Step 2→3 is exactly the dialog-owns-the-only-tooltips case. Confirmed with a throwaway test at your HEAD (fails), and `isKeyboardModality = false` on the `0→1` transition turns it green.

We should probably just drop the refcount though — install once behind `typeof document !== 'undefined'` and never tear down, then the window can't exist. Two listeners for the page lifetime, hook goes 55 → ~28 lines, and `useEffect`/`useCallback` both disappear. Verified green + `tsc` clean if you want the diff.

src/core/components/tooltip/useKeyboardModality.ts
```diff
@@ -0,0 +1,55 @@
...
+                document.removeEventListener('keydown', handleKeyDown, true);
+                document.removeEventListener('pointerdown', handlePointerDown, true);
+            }
+        };
+    }, []);
+
+    return useCallback(() => isKeyboardModality, []);
```
thekidnamedkd: If the refcount goes, this becomes a plain `getIsKeyboardModality()` export — no hook, no `useCallback` (nothing depends on its identity anyway). Also fixes `const isKeyboardModality = useKeyboardModality()` in `tooltip.tsx`, where an `is`-prefixed name is bound to a function.

src/core/components/tooltip/useKeyboardModality.ts
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Reproduces the part of the `:focus-visible` heuristic a tooltip needs: it reports whether the focus that just
+ * happened was driven by the keyboard. The CSS pseudo-class itself cannot be used here, because browsers only expose
+ * it as a style and jsdom aliases it to `:focus`, which would make the distinction untestable.
+ *
+ * Returns a getter rather than a value so that reading it never re-renders the tooltip: the modality is only ever
+ * read from inside an event handler, and the tooltip's own open state drives the render.
+ */
+export const useKeyboardModality = () => {
```
thekidnamedkd: We should give this a `useKeyboardModality.test.ts` — every other hook in the kit has one (`useCopy`, `useDebouncedValue`, `useRandomId`, `useInputProps`, `useNumberMask`). keydown sets true / pointerdown sets false / the remount case above would have caught it.

src/core/components/addressOutput/addressOutput.test.tsx
```diff
@@ -153,6 +156,33 @@ describe('<AddressOutput /> component', () => {
         expect(screen.getByRole('button', { name: 'Copy' })).toBeInTheDocument();
     });
 
+    it.each([
+        { name: 'the reveal button', props: {} },
+        { name: 'the copy control', props: { href: 'https://etherscan.io/address/x' } },
+        { name: 'the copy control only', props: { reveal: false } },
+    ])('keeps every tooltip closed when a dialog autofocuses $name', async ({ props }) => {
```
thekidnamedkd: Since the description calls out `Clipboard` as affected too, worth one of these over there as well — it's covered transitively today, but it'd rot silently.

src/core/components/tooltip/tooltip.test.tsx
```diff
@@ -42,4 +42,44 @@ describe('<Tooltip/> component', () => {
         await user.hover(screen.getByText(trigger));
         await waitFor(() => expect(handleOpenChange).toHaveBeenCalledWith(true));
     });
+
+    it('renders the tooltip content when the trigger is focused with the keyboard', async () => {
```
thekidnamedkd, line 47: Tiny thing: this one passes with or without the gate (pre-fix Radix opened on any focus), so it's over-gating cover rather than one of the "confirmed to fail without the gate" four. Keep it, just not part of that count.

thekidnamedkd (commented):
Nice root-cause writeup, and thanks for pinning down the `composeEventHandlers` ordering — that's the part I'd have gone looking for.

One real bug and a few nits below. On the judgement call you flagged: keep it. Matching native `:focus-visible` is the right call, and the tooltip is the only thing telling a keyboard user what that focused icon button is.

src/core/components/tooltip/keyboardModality.test.ts
```diff
@@ -0,0 +1,44 @@
...
+        expect(getIsKeyboardModality()).toBe(true);
+    });
+
+    it('reports no keyboard modality after a pointer interaction', () => {
+        pressKey();
+        pressPointer();
+        expect(getIsKeyboardModality()).toBe(false);
+    });
+
+    it.each([{ modifier: 'metaKey' }, { modifier: 'altKey' }, { modifier: 'ctrlKey' }])(
```
thekidnamedkd, line 58: `pnpm lint:check` fails here on formatting only — Biome wants the `it.each` array broken across lines. `pnpm lint` fixes it.

src/core/components/tooltip/keyboardModality.test.ts
```diff
@@ -0,0 +1,44 @@
+import { getIsKeyboardModality } from './keyboardModality';
+
+describe('keyboardModality', () => {
+    const pressKey = (init?: KeyboardEventInit) =>
+        document.dispatchEvent(new KeyboardEvent('keydown', { bubbles: true, key: 'Tab', ...init }));
+
+    // jsdom does not implement PointerEvent, and the listener reads nothing off the event.
+    const pressPointer = () => document.dispatchEvent(new Event('pointerdown', { bubbles: true }));
+
+    it('reports no keyboard modality before any interaction', () => {
```
thekidnamedkd: This one only passes because it runs first — `jest --randomize` fails it about a third of the time (the other six are fine). A `beforeEach(pressPointer)` would settle it, or just fold the initial value into the stale-modality test at the bottom.

thekidnamedkd (commented):
All five addressed — and taking the listener-once route means the stale-modality window is gone structurally, not just reset. Nice.

Verified at `91394970`:

- My original repro (tab → unmount → mouse-open → autofocus) now passes.
- Put the mount-tied listeners back: **4 tests fail**. Drop the gate in `tooltip.tsx`: **7 fail**. Both regressions are genuinely covered now.
- Full suite `1240 passed`, `tsc` clean. The one failure is `useFormContext.test.tsx`, which fails on `main` too — not yours.

Two small things left, then good to go from me.

src/core/components/tooltip/keyboardModality.test.ts
milosh86: fixed

src/core/components/tooltip/keyboardModality.test.ts
milosh86: not sure about this, I can't find the failing CI job ... maybe you tested with a different Biome version?

src/core/components/tooltip/tooltip.tsx
```diff
@@ -81,16 +82,59 @@ export const Tooltip: React.FC<ITooltipProps> = (props) => {
         ...otherProps
     } = props;
 
+    const [uncontrolledOpen, setUncontrolledOpen] = useState(defaultOpen ?? false);
+    const isControlled = open != null;
+    const isOpen = isControlled ? open : uncontrolledOpen;
```
thekidnamedkd, line 87: `Tooltip` owns the controlled/uncontrolled switch now, and nothing covers it — `defaultOpen` and `open` have no tests either side of this PR, so the gate could start swallowing a legit open and the suite would stay green. I checked all three paths by hand and they're fine; worth pinning two of them (`defaultOpen` renders, controlled `open={false}` stays shut but still calls `onOpenChange(true)`) so the next change to `handleRootOpenChange` can't quietly break consumers.

thekidnamedkd (approved):
**Retracting my lint comment — that was my fault.** My local `@biomejs/biome` is 2.5.0, the repo wants `^2.5.8`, and 2.5.0 formats `it.each` differently. Re-ran with 2.5.8 pinned: clean on this head *and* on `91394970`, so there was never anything to fix. Sorry for the detour.

The rest checks out at `6d181df5`:

- Order dependence gone — `jest --randomize` on the tooltip dir, 4 runs, `24 passed` each. `isolateModulesAsync` is the right tool.
- Full suite `1255 passed`, `tsc --noEmit` clean, biome 2.5.8 clean. Only failure is `useFormContext.test.tsx`, which fails on `main` too.
- Spot-checked the state rewrite by hand: `defaultOpen`, controlled `open={true}`, and controlled-`open={false}`-still-reports-`onOpenChange(true)` all behave. Close-on-pointer-leave doesn't fire under jsdom, but it doesn't on `main` or `91394970` either — Radix's hoverable-content path, not yours.

The `preventDefault` → refuse-the-open switch is the right trade, and the `Dropdown.Item` test is a good way to pin why. Agreed on the deliberate divergence from native `:focus-visible` too — the description makes the case better than my earlier "keep native parity" take did.

One nit below, non-blocking. Nice work on this.

